PR #12, opened 2017-10-04
Fix large pages on Windows and make them more user-friendly

### LINKED ISSUE #4: active self test
All old miner versions contains a self test of the cpu miner. The code is currently not called and returned always true.

# Todo

- [ ] Activate and test the *self test* functionality.

fireice-uk:
- Add large page initialisation on Windows
- Add a proper UAC manifest requesting elevation when we start
- Fix #4 self_test() - this is integral part of initialisation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -401,6 +401,10 @@ set(LIBRARY_OUTPUT_PATH "bin")
 
 target_link_libraries(xmr-stak ${LIBS} xmr-stak-c xmr-stak-backend)
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+    set_property(TARGET xmr-stak PROPERTY LINK_FLAGS "/level='requireAdministrator' /uiAccess='false'")
+endif()
+
 ################################################################################
 # Install
 ################################################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -401,10 +401,6 @@ set(LIBRARY_OUTPUT_PATH "bin")
 
 target_link_libraries(xmr-stak ${LIBS} xmr-stak-c xmr-stak-backend)
 
-if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -401,6 +401,10 @@ set(LIBRARY_OUTPUT_PATH "bin")
 
 target_link_libraries(xmr-stak ${LIBS} xmr-stak-c xmr-stak-backend)
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+    set_property(TARGET xmr-stak PROPERTY LINK_FLAGS "/level='requireAdministrator' /uiAccess='false'")
```
psychocrypt: Does this mean that the miner must run always started as administrator under windows?

fireice-uk:
@psychocrypt I hacked around UAC in a different way. Now the miner will run no matter if the user presses yes or no, however if he presses yes it will run in a different process than original.

Middle ground is impossible here because UAC would be too easy to bypass if you could elevate a running process.

psychocrypt:
Is it really needed that the UAC always pops up? Or do wee need it only
once and than large pages is activated on the system.
In the case we need it only one we should think about to it only if no
config.txt is available or build a extra little helper tool which can but
not must executed by the user to activate large pages.

Am 05.10.2017 12:17 Vorm. schrieb "fireice-uk" <notifications@github.com>:

@psychocrypt <https://github.com/psychocrypt> I hacked around UAC in a
different way. Now the miner will run no matter if the user presses yes or
no, however if he presses yes it will run in a different process than
original.

Middle ground is impossible here because UAC would be too easy to bypass if
you could elevate a running process.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/fireice-uk/xmr-stak/pull/12#issuecomment-334305553>,
or mute
the thread

fireice-uk:
@psychocrypt UAC prompt is needed if you want to allocate large pages, so every time unfortunately, unless the uses only wants to use slow memory.

psychocrypt:
:-( I will try you changes tomorrow, now it's to late

Issue #5, opened 2017-10-03
OpenCL name of a VEGA GPU


psychocrypt:
The auto suggestion for the AMD backend will add the name of the GPU as comment into the config.
What is the name of the VEGA gpu provided by OpenCL?

**Could someone from the communityplease  provide us with the config created by the auto suggestion!**

This information is needed to improve the auto suggestion for VEGA gpus.

anhphan:
I use self build on windows with only one custom offtion `-DCUDA_ARCH=50;61` because cuda 9 drop computed 2.0

Just tested on server with RX Vega 56 (blockchain driver) and GTX 1050Ti (only install cuda 8 runtime - no cuda 9)

xmr-stak auto disable AMD backend so cannot give you the name it generated.

Don't know how to force it enable.
```
-------------------------------------------------------------------
WARNING: NVIDIA Insufficient driver!
[2017-10-05 00:05:57] : WARNING: backend NVIDIA disabled.
[2017-10-05 00:05:57] : Found AMD platform index id = 1, name = Advanced Micro Devices, Inc.
[2017-10-05 00:05:57] : WARNING: backend AMD disabled.
[2017-10-05 00:05:57] : CPU configuration stored in file 'cpu.txt'
```

psychocrypt:
The opencl library from the installed cuda is selected instead your amd
opencl lib.
Set the path to the amd libopencl.so

`LD_LIBRARY_PATH=PATHTOOPENCL:$LD_LIBRARY_PATH ./xmr-stak`

you can find the path to amd opencl with `find /usr | grep -i libopencl`

Am 04.10.2017 7:13 Nachm. schrieb "Anh Phan Tuan" <notifications@github.com

anhphan:
It's windows server 2016, not linux, and platform index 1 is correct index for AMD (NVIDIA is 0)

psychocrypt:
ignore what I am wrote before. It could be a bug that the autosuggestion stops of a non amd gpu is found. I need to check it today.

Please compile with the cmake option `-DCUDA_ENABLE=OFF` mabe this helps to activate the amd gpu.

anhphan:
AMD can be active if i manual created amd.txt file with thread and platform info copy from old amd version, it's just autosuggestion not work, (and large page not work too)

psychocrypt:
Could you please open an issue for the large page support and descriibe if
you mean the option within config.txt or how you see that is not working.

[...]

PR #13, opened 2017-10-04
fix amd auto suggestion

psychocrypt:
Platform index zero was always used to search for the device detection.

Fix: use correct detected AMD platform index

Fix bug detected from @anhphan https://github.com/fireice-uk/xmr-stak/issues/5#issuecomment-334225424

### DIFF
--- a/xmrstak/backend/amd/autoAdjust.hpp
+++ b/xmrstak/backend/amd/autoAdjust.hpp
@@ -52,14 +52,16 @@ public:
 			return false;
 		}
 
-		devVec = getAMDDevices(0);
+		devVec = getAMDDevices(platformIndex);
 
 
 		int deviceCount = devVec.size();
 
 		if(deviceCount == 0)
-            return false;
-
+		{
+			printer::inst()->print_msg(L0,"WARNING: No AMD device found.");
+			return false;
+		}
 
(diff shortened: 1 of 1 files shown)


xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -52,14 +52,16 @@ class autoAdjust
...
-		devVec = getAMDDevices(0);
+		devVec = getAMDDevices(platformIndex);
 
 
 		int deviceCount = devVec.size();
 
 		if(deviceCount == 0)
+		{
+			printer::inst()->print_msg(L0,"WARNING: No AMD device found.");
             return false;
```
fireice-uk: Can you fix that indent?

psychocrypt: Done, after we added the most changes I will run `sed` over the code to remove all my spaces with taps.

fireice-uk: Thanks, keep in mind that is is much easier to go the other way round (sed s/\t/    /g), one of the reasons why I always use tabs. 

When changing spaces into tabs make sure to include ^ anchor to prevent random tabs in the code.

fireice-uk:
CI passed, merging.

Issue #2, opened 2017-10-03
check nicehash pool support


psychocrypt:
During the refactoring I have remove the nicehash pool support, I added it back with https://github.com/fireice-uk/xmr-stak/commit/1f40f88a0c031db1bb04fa39a51117263eb27155.

I have not tested if it runs without issues, therefore a runtime test is needed.

- [ ] do test against the nicehash pool.

anhphan:
Nicehash is totally broken, I try with both cpu backend and amd backend, both receiver 100% reject with error on the web page:

`Invalid nonce; is miner not compatible with NiceHash?`

psychocoderHPC:
Could you provide me with your config.txt thus I can try to mine against your nicehash account to fix the issue.

psychocrypt:
Could you provide me with a config.txt which I can use for testing. I will spend my hash power to your account to fix it.

anhphan:
The config file is attached.

[config.txt](https://github.com/fireice-uk/xmr-stak/files/1356859/config.txt)

psychocrypt:
@anhphan Thx for providing me with a account, the nice hash support is fixed in #14

anhphan:
It's fixed somewhere but it's not completely fix.
Will try more with each backend (cpu, amd). This is my result when running both amd and cpu (5.5kh amd + 0.9kh cpu)

![image](https://user-images.githubusercontent.com/1614690/31198367-3b0afc8e-a97e-11e7-9f8c-aa03ba9f82c5.png)

anhphan:
Seem good (with shift 8 bit as @fireice-uk indicated in review code).
45 min mining on nicehash (nh will push us very high diff if we have high speed, then maybe problem out of nonce will not happen):
![image](https://user-images.githubusercontent.com/1614690/31201525-2f2f72c2-a989-11e7-9980-1edffb76a35c.png)

fireice-uk:
@anhphan thanks for testing this one! Closing as the fix was merged in #14

PR #14, opened 2017-10-04
fix broken nicehash support

psychocrypt:
- add nicehash flag back to `miner_work`

This bug was detected in #2

### DIFF
--- a/xmrstak/backend/miner_work.hpp
+++ b/xmrstak/backend/miner_work.hpp
@@ -26,8 +26,8 @@ namespace xmrstak
 		miner_work() : iWorkSize(0), bNiceHash(false), bStall(true), iPoolId(0) { }
 
 		miner_work(const char* sJobID, const uint8_t* bWork, uint32_t iWorkSize, uint32_t iResumeCnt,
-			uint64_t iTarget, size_t iPoolId) : iWorkSize(iWorkSize), iResumeCnt(iResumeCnt),
-			iTarget(iTarget), bNiceHash(false), bStall(false), iPoolId(iPoolId)
+			uint64_t iTarget, bool bNiceHash, size_t iPoolId) : iWorkSize(iWorkSize), iResumeCnt(iResumeCnt),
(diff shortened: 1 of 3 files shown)


xmrstak/backend/iBackend.hpp
```diff
@@ -38,7 +38,7 @@ namespace xmrstak
 		// Limited version of the nonce calc above
 		inline uint32_t calc_nicehash_nonce(uint32_t start, uint32_t resume)
 		{
-			return start | ( calc_start_nonce(resume) >> 8u );
+			return start | ( calc_start_nonce(resume) >> 7u );
```
fireice-uk: Why 7 here? If you want to have the top 8 bits for nicehash and 24 for us, this should be 8.

psychocrypt: Nicehash blocks the highest 8bit within the 32bit nonce. The `calc_start_nonce` function sets for the thread one the bit 31, by shifting the 31th activated bit 7 places right we end with an activated bit in place 23. This means we use bit [0;23] for the nonce.

fireice-uk: Still not quite getting it here, let's talk numbers. Is calc_start_noce using everything up to the MSB? So for example 0xFFFF0000 - or does it leave it blank - 0x7FFF0000?

fireice-uk: Just to give you food for thought:
0xAC000000 | (0xFFFF0000 >> 7) = 0xADFFFE00

psychocrypt: I fixed it, thx @fireice-uk

anhphan: Yes 7-bit shifting make about 50% share rejected (on cpu it's duplicated share, on gpu it's is wrong nonce).

Try rebuild with 8-bit shifting value, everything seem good now.

Let's it run a bit longer to check do we have out of nonce problem with high hashrate.

fireice-uk:
@anhphan Can you verify it is working?

anhphan:
It's work fine @fireice-uk; forgot to update status here and already update in issue #2.

PR #16, opened 2017-10-04
fix invalid kernel call parameter

### LINKED ISSUE #1: Error CL_INVALID_WORK_GROUP_SIZE
The error `Error CL_INVALID_WORK_GROUP_SIZE when calling clEnqueueNDRangeKernel for kernel 0.` still exists and was described in issue https://github.com/fireice-uk/xmr-stak-amd/issues/93.

This is a reminder for me, I already found the reason for this bug.

psychocrypt:
The number of threads within a kernel must be a multiple of the worksize.
If not it can crash on some systems.

fix #1

# Tests

- [x] runtime test ~~(please do not merge before the tests passed)~~

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/gpu.cpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.cpp
@@ -653,6 +653,8 @@ size_t XMRSetJob(GpuContext* ctx, uint8_t* input, size_t input_len, uint32_t tar
 
 	input[input_len] = 0x01;
 	memset(input + input_len + 1, 0, 88 - input_len - 1);
+	
+	size_t numThreads = ctx->rawIntensity;
 
 	if((ret = clEnqueueWriteBuffer(ctx->CommandQueues, ctx->InputBuffer, CL_TRUE, 0, 88, input, 0, NULL, NULL)) != CL_SUCCESS)
 	{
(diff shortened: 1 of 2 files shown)


psychocrypt:
Yes will do it with an cassert.

Am 05.10.2017 12:45 Nachm. schrieb "fireice-uk" <notifications@github.com>:

psychocrypt:
I added the assert

psychocrypt:
I can also add a normal if, but due to the up round above it can only fail if someone change the code, in that case the assert hopefully is showing it during the tests before the release.

But if you think that there must be an secure check I add it.

fireice-uk:
@psychocrypt Ah, I missed the roundup, I'm not quite sure about it though. I **really** dislike when user tells us one thing and we do another. Is is possible to add a warning there?

psychocrypt:
Do you mean a warning that the intensity must be a multiple of the work
group size.
The user input is and was correct and we operate only on as many threads a
the user defined but to start the algorithm there is a library(opencl)
limitation.
To limit the intensity to a multiple would cost performance.
I think of changing the behaviour for nvidia also because currently we
waste performance,by using blocks and threads in cuda we always have a
multiple but that cost use a few percent of the performance.

Am 06.10.2017 00:04 schrieb "fireice-uk" <notifications@github.com>:

fireice-uk:
Ah. Got it now. Intensity and global threads were equivalent until now. In that case can you just confirm that the code from lines 248 - 288 is still correct?

psychocrypt:
The lines are fine.
The number of threads (intensity) used for hashing is not changed therefore
the amount of memory used is equal to the current dev version.

PR #24, opened 2017-10-06
remove early returns

psychocrypt:
Early returns within a kernel can have unexpected behavior, this strongly depends on the opencl
runtime compiler. To avoid errors all early returns are removed.
There is no negative effect(e.g.performance) if the threads stay alive up to the end of the kernel.

One source for the early return is: http://al-key-opencl.blogspot.de/2014/09/be-careful-not-to-mix-early-return-and.html

This pull request also fix a race condition where all fill the shared memory.
Also a bug introduced with #16 is fixed, because of the early return not was possible that
the last block works with an wrong initilized shared memory (result should be a wrong hash if the result target size is valid).

# Review

To review this PR please use the [diff without withe space changes](https://github.com/fireice-uk/xmr-stak/pull/24/files?w=1) this will help to the the important  code changes.

# Test
- [x] runtime test
- [x] must be rebased against #23 to avoid merge conflicts

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
+++ b/xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
@@ -422,12 +422,10 @@ __kernel void cn0(__global ulong *input, __global uint4 *Scratchpad, __global ul
 	uint4 text;
 
 	const ulong gIdx = get_global_id(0) - get_global_offset(0);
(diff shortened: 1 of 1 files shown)


fireice-uk:
Looks a-ok, can you rebase this one?

psychocrypt:
PR is rebased

PR #27, opened 2017-10-08
improve AMD auto suggestion

psychocrypt:
- use always a multiple of the number of compute units
- add more device information to the config file

On my Pitcairn GPU I got a 5% performance improvement. The improvement depends on how far the old suggestion is from a multiple of the number of compute units.


xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -79,6 +79,8 @@ class autoAdjust
 		configEditor configTpl{};
 		configTpl.set( std::string(tpl) );
 
+		constexpr size_t byte2mib = 1024u * 1024u;
```
fireice-uk: Did you mean 2MiB or 1MiB here? Either the name or the value doesn't match.

psychocrypt: good point, I will change it to `byteToMiB`

psychocrypt:
I change to annotated point

xmrstak/backend/amd/autoAdjust.hpp
anhphan: Just personally, 

But for all const like this, i always use naming style:
`<First Unit Name> Per [Number] <Second Unit Name>`

Sample:
`SecondPerMinute = 60`
`SecondPerDay = 24 * 3600`
`SecondPer5Day = 5 * 24 * 3600`
`BytePerMByte = 1024 * 1024`
`BytePer2MByte = 2 * 1024 *1024`

PR #33, opened 2017-10-10
fix setaffinity

### LINKED ISSUE #22: affinity randomly wrong
I tested the thread pinning under linux. Something went sometimes wrong.
Normally I use thread 0 and 1 on my laptop. To get a higher chance to trigger the bug I configuread it to 0, 3. 
Than I started the miner and looked to `htop`. In 20 -50% the second thread is going to core 1 instead of core 3.

**It looks like a race condition somewhere.**

I checked if hwloc is a problem but also without hwloc I get this behavior. 
~~I also tried to pin the thread with hwloc but with the same result.~~

psychocrypt:
fix #22 

- move setaffinity call out of the constructor
- remove guard variable

After this fix the affinity is always correct in all my tests.

# Test
- [x] runtime test

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ option(CMAKE_LINK_STATIC "link as much as possible libraries static" OFF)
 
 option(CUDA_ENABLE "Enable or disable CUDA support (NVIDIA backend)" ON)
 if(CUDA_ENABLE)
-    find_package(CUDA 7.5 QUIET)
+    find_package(CUDA 7.5)
 
     if(CUDA_FOUND)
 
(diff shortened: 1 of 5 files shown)


fireice-uk:
@psychocrypt I'm uncomfortable with this fix - it has more undefined behaviour than the last one. In particular `bindMemoryToNUMANode` can execute before or after `thd_setaffinity`

Can you submit an alternative PR which sticks to original except that it swaps the order of `thd_setaffinity` and `bindMemoryToNUMANode` inside of `pin_thd_affinity`? I want to exclude that as a possible cause.

psychocrypt:
Memory pinning is independent of the pthread pinning because hwloc gets a
thread id explicit, therefore it is not importent to have a defined order.
I also tried to change the order of numa pinning and thread pinning, that
not solve the issue.
The issue is also visible if I disable hwloc.

I invested many hours into that issu and this is currently the only
solution which solves the issu.

fireice-uk:
> I invested many hours into that issu and this is currently the only
> solution which solves the issu.

This is not a healthy attitude. Programming is an iterative process, if you view anything except the first draft as a waste of time then your code will always look like .... first draft.

fireice-uk:
I screwed up slightly (pushed some extra stuff from the working tree), but the PR looks still mergable.

Can you test this one and the alternative on the affected system? They work the same on mine.

fireice-uk:
@psychocrypt Can you review the fix A, B and C and tell me which one works? They all have well defined order of calls. 

If fix B and C work, then the issue is about setting the affinity from this thread. 

If C and A or only B work then the issue is about call order.

PR #36, opened 2017-10-11
Affinity fix version A

fireice-uk:
Order of calls in this PR
1. thd_setaffinity, in this thread
2. bindMemoryToNUMANode


psychocrypt:

fireice-uk:
Tested under Windows - doesn't work. Rejecting this one.

PR #42, opened 2017-10-12
Affinity fix version B

fireice-uk:
Order of calls in this PR
1. thd_setaffinity, not in this thread
2. bindMemoryToNUMANode


psychocrypt:

psychocrypt:
Could you use the some mechanisms to set the affinity for the back-ends amd and nvidia

fireice-uk:
@psychocrypt This one fails to set affinity on Windows too.

psychocrypt:
Ahh I tested only linux :(

PR #43, opened 2017-10-12
Affinity fix version C

fireice-uk:
Order of calls in this PR:

1. bindMemoryToNUMANode
2. thd_setaffinity, not in this thread

### DIFF
--- a/xmrstak/backend/cpu/minethd.cpp
+++ b/xmrstak/backend/cpu/minethd.cpp
@@ -50,22 +50,8 @@
 #include <thread>
 #include <bitset>
 
-
 #ifdef _WIN32
 #include <windows.h>
-
-namespace xmrstak
-{
-namespace cpu
-{
-void minethd::thd_setaffinity(std::thread::native_handle_type h, uint64_t cpu_id)
-{
-	SetThreadAffinityMask(h, 1ULL << cpu_id);
-}
-
-} // namespace cpu
-} // namespace xmrstak
-
 #else
 #include <pthread.h>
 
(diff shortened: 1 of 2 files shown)


psychocrypt:

fireice-uk:
Looks like for Windows to be happy the order of calls needs to be like in this patch. 

I find it puzzling however that all of them would work. Can you give this one a bit more testing on the systems that were affected on your side?

psychocrypt:
Yes I will look deeper to this PR, but could be not before end of the next
week.

xmrstak/backend/cpu/minethd.cpp
```diff
@@ -317,23 +308,17 @@ minethd::cn_hash_fun minethd::func_selector(bool bHaveAes, bool bNoPrefetch)
...
-#if defined(__APPLE__)
-	printer::inst()->print_msg(L1, "WARNING on MacOS thread affinity is only advisory.");
-#endif
-	thd_setaffinity(oWorkThd.native_handle(), affinity);
 }
 
 void minethd::work_main()
 {
 	if(affinity >= 0) //-1 means no affinity
 		pin_thd_affinity();
```
psychocrypt: `bindMemoryToNUMANode(affinity);` should be called here directly and `pin_thd_affinity` can be removed.

fireice-uk: Ok. I will test that one on Windows too.

fireice-uk:
I think this one is done. Let me know.

PR #44, opened 2017-10-12
Make sure all singletons are set to null and make env a global ptr

fireice-uk:
This change makes sure of two things:
- Singleton ptr is always null at first
- We have only single env instance, so we don't duplicate singletons if we create them after the dll is loaded

Not testes at all, but should work. Let me know if there are any problems

### DIFF
--- a/xmrstak/misc/environment.hpp
+++ b/xmrstak/misc/environment.hpp
@@ -12,35 +12,30 @@ class params;
 
 struct environment
 {
+	static inline environment& inst(environment* init = nullptr)
+	{
+		static environment* env = nullptr;
 
-	static environment& inst()
-	{
-		static environment env;
-		return env;
-	}
-	
-	environment& operator=(const environment& env)
-	{
-		this->pPrinter = env.pPrinter;
-		this->pglobalStates = env.pglobalStates;
-		this->pJconfConfig = env.pJconfConfig;
-		this->pExecutor = env.pExecutor;
-		this->pParams = env.pParams;
-		return *this;
+		if(env == nullptr)
+		{
+			if(init != nullptr)
+				env = new environment;
+			else
+				env = init;
+		}
+
+		return *env;
 	}
 
-
-	environment() : pPrinter(nullptr), pglobalStates(nullptr)
+	environment()
 	{
 	}
 
-
-	printer* pPrinter;
-	globalStates* pglobalStates;
-	jconf* pJconfConfig;
-	executor* pExecutor;
-	params* pParams;
-
+	printer* pPrinter = nullptr;
+	globalStates* pglobalStates = nullptr;
+	jconf* pJconfConfig = nullptr;
+	executor* pExecutor = nullptr;
+	params* pParams = nullptr;
 };
 
 } // namepsace xmrstak
--- a/xmrstak/misc/environment.hpp
+++ b/xmrstak/misc/environment.hpp
@@ -18,7 +18,7 @@ struct environment
 
 		if(env == nullptr)
 		{
-			if(init != nullptr)
+			if(init == nullptr)
 				env = new environment;
 			else
 				env = init;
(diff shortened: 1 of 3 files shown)


xmrstak/misc/environment.hpp
```diff
@@ -12,35 +12,30 @@ class params;
 
 struct environment
 {
-
-	static environment& inst()
-	{
-		static environment env;
```
psychocoderHPC: Normally this is the comon way to create a singleton and is free of memory leeks. Is there a reason why you have changed it?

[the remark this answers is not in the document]
fireice-uk: This is because you are creating a copy in the DLL. If you create a singleton after you create a DLL copy you will have a duplicate.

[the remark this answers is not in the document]
fireice-uk: @psychocrypt Can you let me know if you approve or not here?

psychocrypt:
I approved this PR but can't test it currently in windows. Do you tested it in windows?

fireice-uk:
There is no OS-specific code in that PR - merging.

psychocrypt:
note: all the singleton registering in one instance is only made for
windows. Shared library handling dis different in windows and linux :-(

fireice-uk:
What's the difference? There is none from what I can see in the code.

psychocrypt:
In the current implementation there is no difference. But the reason why we need to path the singleton class environment to the dll/so is that a singleton under windows is only unique within a shared libarary. This means if the main binary create a singleton the singleton is not visible in the dll. In linux a singleton is which is created in the shared library or in the main binary is unique ofer all shared libraries and the main binary.
There are whys in the microsoft knowlage base collection where some ways are descriped how you can achieve it under windows but none of this worked in my tests.
That's the reason why all all singleton pointer are living in the environment class.

fireice-uk:
You mean the difference in symbol visibility between Windows and Linux? If not then please link to the MSDN article.

psychocrypt:
> You mean the difference in symbol visibility between Windows and Linux?

Yes, for me as linux developer this was absolutely new, that a singleton is not unique because of the difference in the symbol export handling.

Sry I can't point to the sources I worked on this during the refactoring and have not stored it :-( 

note to myself: bookmark links

xmrstak/misc/environment.hpp
```diff
@@ -12,35 +12,30 @@ class params;
...
+	{
+	}
 
-	printer* pPrinter;
-	globalStates* pglobalStates;
-	jconf* pJconfConfig;
-	executor* pExecutor;
-	params* pParams;
-
+	printer* pPrinter = nullptr;
```
psychocrypt: @fireice-uk Could you please point me to a reference which describe why it is allowed to initialize member out of the constructor. Is this because of initializer lists?

psychocrypt: Ahh http://en.cppreference.com/w/cpp/language/data_members

xmrstak/misc/environment.hpp
```diff
@@ -12,35 +12,30 @@ class params;
...
-		this->pExecutor = env.pExecutor;
-		this->pParams = env.pParams;
-		return *this;
-	}
+		static environment* env = nullptr;
 
+		if(env == nullptr)
+		{
+			if(init == nullptr)
+				env = new environment;
```
psychocrypt: Could it be that this is the reason for #51. That environment is handled as POD and the pointer will not be initilized with a `nullptr`?

psychocrypt: https://stackoverflow.com/a/15212351

psychocrypt: If this can result in a uninitialized type than we need also check all other singletons.

xmrstak/misc/environment.hpp
fireice-uk: That's correct. For C++11 both those styles are equivalent. I chose the second one because it is clearer and nullptr is a constexpr.

PR #47, opened 2017-10-14
Re-add UAC manifest and WIN_UAC cmake option

fireice-uk:
Thanks for help @anhphan This version includes a UAC manifest by default, since Windows 7 insists on running elevated to lock large pages.

### DIFF
--- a/doc/compile.md
+++ b/doc/compile.md
@@ -44,6 +44,9 @@ After the configuration you need to compile the miner, follow the guide for your
   - there is no *http* interface available if option is disabled: `cmake .. -DMICROHTTPD_ENABLE=OFF`
 - `OpenSSL_ENABLE` allow to disable/enable the dependency *OpenSSL*
   - it is not possible to connect to a *https* secured pool if option is disabled: `cmake .. -DOpenSSL_ENABLE=OFF`
+- `WIN_UAC` will enable or disable the "Run As Administrator" prompt on Windows.
+  - UAC confirmation is needed to use large pages on Windows 7.
+  - On Windows 10 it is only needed once to set up the account to use them.
 
 ## CPU Build Options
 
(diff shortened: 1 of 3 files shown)


doc/compile.md
```diff
@@ -44,6 +44,9 @@ After the configuration you need to compile the miner, follow the guide for your
   - there is no *http* interface available if option is disabled: `cmake .. -DMICROHTTPD_ENABLE=OFF`
 - `OpenSSL_ENABLE` allow to disable/enable the dependency *OpenSSL*
   - it is not possible to connect to a *https* secured pool if option is disabled: `cmake .. -DOpenSSL_ENABLE=OFF`
+- `WIN_UAC` will enable or disable the "Run As Administrator" prompt on Windows.
+  - UAC confirmation is needed to use large pages on Windows 7.
+  - On Windows 10 it is only needed once to set up the account to use them.
```
psychocrypt: Can we compile by default a small additional programm to activate large page support for windows and linux.
In Win10 the PR workflow is

- start the miner once
- edit the config and enable the option
- start miner as admin 
- change config back (for security reason)
- start miner

With a small helper the workflow is:

- start the helper as admin (only once)
- run the miner ( withou admin)

Never the less I would keep the option in the config.

psychocrypt: But to avoid the issue with the ignored command line parameter switching to a seperate programm to activate huge page support can be easily solves the issues.

fireice-uk: On Win7  large pages still require admin rights. That's the main problem. I'm not willing to have two versions **and** a separate program.

psychocrypt:
Maybe we can add the UAC manifest within a dll (equal to the backends) and
call this dll only if it is not possible the allocate large pages. This
means that the dialog is in win10 only triggered once and in win7 always.

fireice-uk:
I will do some testing whether that will work. Don't think so, because that could be used to elevate during runtime and thus bypass UAC.

fireice-uk:
I tested and confirmed that dll's manifest is disregarded, and non-elevated process will load a non-elevated dll. Not really surprising. 

Given the security issues involved my inclination is to merge this one as-is and let users compile a non-uac version.

PR #50, opened 2017-10-17
Make hwloc check capabilities and stop it from leaking mem

fireice-uk:


fireice-uk:
CI passed at last, merging

Issue #52, opened 2017-10-18
Autosuggest issues and fine tuning


gboddin:
Hi !

First, uber nice work on this all-in-one refactoring, thanks

I gave it a shot yesterday on debian sid + cuda 9.0 and hashrates seems more stable than 5 separate miners !

Machine A : 
- Intel(R) Core(TM) i3-7100T CPU @ 3.40GHz ( 2x30H/s, but expected, who cares :) )
- rx580 4GB ( 770H/s )
- gtx 1060 3GB ( 520/H/s )

Machine B (laptop): 
- Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz ( 3x60H/s, that's always nice )
- gtx 950m 4GB ( 170H/s )

I had to adapt the autosuggest on the first machine for the 1060, first it gave : 
```
  { "index" : 0,
    "threads" : 52, "blocks" : 27,
    "bfactor" : 0, "bsleep" :  0,
    "affine_to_cpu" : false,
  },
```

but this was giving ~10% rejected shares ( at 520/Hs ), dividing by 2 

```
  { "index" : 0,
    "threads" : 26, "blocks" : 27,
    "bfactor" : 0, "bsleep" :  0,
    "affine_to_cpu" : false,
  },
```
made all the rejects go away and we're now at 100% accepted for 12h :)

Hope it helps !

psychocrypt:
Thanks for opening the issue. Could you please repeat the test and post the
reason for the rejected shares. If the miner gets a new block and just
found e.g. 4 shares for the old block than it could be that you get 4
expired shares but this can not be avoided.
If you get compute errors than we need to look deeper to found the reason.

fireice-uk:
@gboddin can you confirm that the testing was on topic-nonce-final branch?

gboddin:
It was without the topic-nonce-final.

I'm merging, building and will report again

gboddin:
Ok first report on 950m : 
Letting autoconfig gives : 

```
"gpu_threads_conf" :
[
  // gpu: GeForce GTX 950M architecture: 50
  //      memory: 4000/4044 MiB
  { "index" : 0,
    "threads" : 68, "blocks" : 15,
    "bfactor" : 0, "bsleep" :  0,
    "affine_to_cpu" : false,
  },

],
```
however, this ends up with : 

```
-------------------------------------------------------------------
xmr-stak 2.0.0-predev mining software.
[...]

Issue #34, opened 2017-10-11
GPU ERROR issue [possibly CUDA 7.5]


psychocrypt:
This text is copied out of https://github.com/fireice-uk/xmr-stak/pull/28#pullrequestreview-68650125

The compile still fails, but we are making progress. I googled for a fix and adding `add_definitions(-D_MWAITXINTRIN_H_INCLUDED)` seems to fix the problem.

```
/usr/lib/gcc/x86_64-linux-gnu/5/include/mwaitxintrin.h(36): error: identifier "__builtin_ia32_monitorx" is undefined

/usr/lib/gcc/x86_64-linux-gnu/5/include/mwaitxintrin.h(42): error: identifier "__builtin_ia32_mwaitx" is undefined

2 errors detected in the compilation of "/tmp/tmpxft_00007a1d_00000000-22_cuda_extra.compute_20.cpp1.ii".
CMake Error at xmrstak_cuda_backend_generated_cuda_extra.cu.o.cmake:266 (message):
  Error generating file
  /home/fireice/github/xmr-stak/build/CMakeFiles/xmrstak_cuda_backend.dir/xmrstak/backend/nvidia/nvcc_code/./xmrstak_cuda_backend_generated_cuda_extra.cu.o
```

Can you make sure if that flag needs to be targeted at GCC or not.

-----------------------------------------------------
added by psychocrypt: This error will only occur if the unsupported compiler gcc 5.X is enabled by manually editing of the file `host_config.h`

psychocrypt:
@fireice-uk Could it be that you changed your CUDA headers? Normally CUDA 7.5 not support gcc 5.X and should crash with the error message

```
include/host_config.h:115:2: error: #error -- unsupported GNU version! gcc versions later than 4.9 are not supported!
 #error -- unsupported GNU version! gcc versions later than 4.9 are not supported!
```

psychocrypt:
fixed with #35

ostap-mykhaylyak:
"Same" problem with CUDA with compile on Windows.

CPU and OPENCL work but not compile cuda ddl

fireice-uk:
@ostap-mykhaylyak Can you confirm that you can compile ok now?

ostap-mykhaylyak:
No compile error, but when run CUDA crash.

[CUDA] Error gpu 0: <C:/xmr-stak-dev/xmrstak/backend/nvidia/nvcc_code/cuda_extra.cu>:214

[...]

PR #57, opened 2017-10-20
fix illegal memory access

psychocrypt:
- remove restricted pointer
- change some sizes to `unit32_t` to avoidwrong compiler optimization

The blake algorithm code is equal to the host code implementation but the developer added to all pointer parameter the `restrict` keyword. The usage of restrict can have negative effect because the compiler is free to do many code optimizations. This can result in errors. I identified the broken line and checked why the illegal memory access was triggered. I can't find any bug expected the `restrict` keyword.
By changing some array sizes to `uint32_t` another error can be removed. It could be that this is an compiler bug (this is not the first time that some `int` optimizations created wrong code).

The bug is not triggered on all gpus (I got it with sm_70).

This PR solves closed many issues from xmr-stak-nvidia.

This PR is not increasing the register usage or the spill loads/stores.

~~**- [x] please do not merge maybe this is only workaround sometimes the issue, I am currently testing other fixes**~~

### DIFF
--- a/xmrstak/backend/nvidia/nvcc_code/cuda_blake.hpp
+++ b/xmrstak/backend/nvidia/nvcc_code/cuda_blake.hpp
@@ -2,7 +2,8 @@
 
 typedef struct {
 	uint32_t h[8], s[4], t[2];
-	int buflen, nullt;
+	uint32_t buflen;
+	int nullt;
 	uint8_t buf[64];
 } blake_state;
 
@@ -50,7 +51,7 @@ __constant__ uint32_t d_blake_cst[16]
 	0xC0AC29B7, 0xC97C50DD, 0x3F84D5B5, 0xB5470917
 };
 
-__device__ void cn_blake_compress(blake_state * __restrict__ S, const uint8_t * __restrict__ block)
+__device__ void cn_blake_compress(blake_state *  S, const uint8_t *  block)
 {
 	uint32_t v[16], m[16], i;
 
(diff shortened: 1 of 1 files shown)


psychocrypt:
This PR maybe solves #52 and #34

psychocrypt:
This pull request is now tested again. I found a way to trigger the issue on different GPUs, with this PR the issue is gone on all gpus.

patch to re produce this issue:
```
diff --git a/xmrstak/backend/nvidia/nvcc_code/cuda_extra.cu b/xmrstak/backend/nvidia/nvcc_code/cuda_extra.cu
index 7734473..be8ae0a 100644
--- a/xmrstak/backend/nvidia/nvcc_code/cuda_extra.cu
+++ b/xmrstak/backend/nvidia/nvcc_code/cuda_extra.cu
@@ -132,9 +132,9 @@ __global__ void cryptonight_extra_gpu_final( int threads, uint64_t target, uint3
        uint64_t hash[4];
        uint32_t state[50];
 
-#pragma unroll
-       for ( i = 0; i < 50; i++ )
-               state[i] = ctx_state[i];
+//#pragma unroll
+       //for ( i = 0; i < 50; i++ )
+       //      state[i] = ctx_state[i];
 
        cn_keccakf2( (uint64_t *) state );
```

psychocrypt:
I have not tested it. Therr is no negative effect on the performance or register usage.
Also I think restrict is used wrong, because in one function the restrict keyword was added to paramter `S` and in the same moment a second paramter is passed which points to a member of `S`.

fireice-uk:
@psychocrypt in that case you are correct - restrict pointers should be memcpy-safe. Merging

PR #58, opened 2017-10-20
fix CUDA 9 shuffle warning

psychocrypt:
use `__shffl_snyc` if CUDA 9+ is avalable

warning:
```
xmr-stak/xmrstak/backend/nvidia/nvcc_code/cuda_core.cu(159): warning: function "__shfl(int, int, int)"
cuda_9_0/include/sm_30_intrinsics.hpp(152): here was declared deprecated ("__shfl() is not valid on compute_70 and above, and should be replaced with __shfl_sync().To continue using __shfl(), specify virtual architecture compute_60 when targeting sm_70 and above, for example, using the pair of compiler options: -arch=compute_60 -code=sm_70.")
```

### DIFF
--- a/xmrstak/backend/nvidia/nvcc_code/cuda_core.cu
+++ b/xmrstak/backend/nvidia/nvcc_code/cuda_core.cu
@@ -156,7 +156,11 @@ __forceinline__ __device__ uint32_t shuffle(volatile uint32_t* ptr,const uint32_
 #else
     unusedVar( ptr );
     unusedVar( sub );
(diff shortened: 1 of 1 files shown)


fireice-uk:
@psychocrypt Can you fix the wonky indenting? Each level should be a single \t

psychocrypt:
This file is fully indented by spaces. I would prevent to mix it. 

I will do a full space to tab commit before the release.

fireice-uk:
Ok, thanks.

PR #67, opened 2017-10-24
add Aeon support

psychocrypt:
Add support to mine Aeon and Monero from one miner.

By default the miner is build to support both currencies. It is possible to compile the miner to support Monero or Aeon only.
There is a new command line option `--currency` and a CMake option `XMR-STAK_CURRENCY` added.
It was on purpose that the options are named `currency` and the user can choose between `monero` and `aeon`. From the perspective of a non export users there are only different currencies and it could be complicated for to understand that the algorithm behind differ and one is called cryptonight and the other cryptonight-lite.

# Review

I have split the changes into logical commits thus each commit can be easily review instead of the hole changes. 

# Test

- [x] CPU aeon
- [x] CPU xmr
- [x] NVIDIA aeon
- [x] NVIDIA xmr
- [x] AMD aeon
- [x] AMD xmr

Note: I have currently no AMD device available (my test system is dead :-( ) Therefore I tested the OpenCL AMD part only on a NVIDIA card. **Could some please test this changes with an AMD device.**


xmrstak/backend/cpu/minethd.cpp
```diff
@@ -285,23 +289,43 @@ void minethd::consume_work()
...
+#ifndef CONF_NO_AEON
+		cryptonight_hash<AEON_MASK, AEON_ITER, AEON_MEMORY, false, false>,
+		cryptonight_hash<AEON_MASK, AEON_ITER, AEON_MEMORY, false, true>,
+		cryptonight_hash<AEON_MASK, AEON_ITER, AEON_MEMORY, true, false>,
+		cryptonight_hash<AEON_MASK, AEON_ITER, AEON_MEMORY, true, true>
+#endif
 	};
 
-	std::bitset<2> digit;
+	std::bitset<3> digit;
```
fireice-uk: You misundestood the code above. Please refer to the comment, we are essentially counting in binary with true / false flags.

So:
```
false false   =  0
false true    =  1
true  false   =  2
true  true    = 3
````

For bitset of 3 you need to have 8 entries in the table, and count up to 8.

**Comment by psychocrypt** [comment]
There are 8 value in the table if XMR-STAK_CURRENCY=all.
Else there are 4 values in the table, and bit 3 is always set to 0.

File: xmrstak/cli/cli-miner.cpp [hunk 2]
```diff
@@ -230,10 +244,30 @@ int main(int argc, char *argv[])
 		;
 		configEditor configTpl{};
 		configTpl.set(std::string(tpl));
+		std::cout<<"Please enter:"<<std::endl;
+		auto& currency = params::inst().currency;
+		if(currency.empty())
+		{
+			std::string tmp;
+#if defined(CONF_NO_AEON)
+			tmp = "xmr";
```
fireice-uk: Can you change xmr to proper currency name - Monero?

psychocoderHPC:
I will add some comments to this code part.

psychocrypt:
Yes I can change it to 'monero'. I used 'xmr' to keep it short but you are
right monero could be better.

xmrstak/cli/cli-miner.cpp
psychocrypt: Yes I can change it to 'monero'. I used 'xmr' to keep it short but you are right monero could be better.

fireice-uk:
I hate when that happens too ^^

xmrstak/backend/cpu/minethd.cpp
psychocrypt: I will add some comments to this code part.

fireice-uk: Ok I see it now, my bad, this part is ok as-is

psychocrypt: There are 8 value in the table if XMR-STAK_CURRENCY=all. Else there are 4 values in the table, and bit 3 is always set to 0.

psychocrypt:
I changed xmr to monero. The writing of monero is case insensitiv to avoid that we get many support tickets because some are writing `monero` and other `Monero`

psychocrypt:
Could you please point to the code- can't find the singleton in utility.hpp or cpp.

fireice-uk:
https://github.com/fireice-uk/xmr-stak/pull/67/commits/c791cd90cbba4f00aa8877b0c3ad0477fc9bf68f <- this commit added the executor code (+1000 delta lines)

https://github.com/fireice-uk/xmr-stak/pull/67/commits/777935caff91fc2cadd65aae919518e2b476db64 <- removes the code

Can you squash those two together please?

psychocrypt:
Yes I will do.

fireice-uk:
I have an RX480 test rig set up, so will do the AMD testing after the squash

psychocrypt:
I rebased and squash the PR.

xmrstak/cli/cli-miner.cpp
```diff
@@ -230,10 +245,30 @@ int main(int argc, char *argv[])
 		;
 		configEditor configTpl{};
 		configTpl.set(std::string(tpl));
+		std::cout<<"Please enter:"<<std::endl;
+		auto& currency = params::inst().currency;
+		if(currency.empty())
```
fireice-uk: This if is never executed in my testing. After `rm *.txt; ./xmr-stak` the miner goes on to mine aeon without prompting.

fireice-uk: Just to make sure checked cmake output -> `-- Set miner currency to 'monero' and 'aeon'`

xmrstak/cli/cli-miner.cpp
```diff
@@ -250,21 +251,21 @@ int main(int argc, char *argv[])
...
 			std::string tmp;
 #if defined(CONF_NO_AEON)
-			tmp = "xmr";
-#elif defined(CONF_NO_XMR)
+			tmp = "monero";
+#elif defined(CONF_NO_MONERO)
 			tmp = "aeon";
 #endif
-			while(tmp.compare("xmr") != 0 && tmp.compare("aeon") != 0)
+			while(xmrstak::strcmp_i(tmp, "monero") && xmrstak::strcmp_i(tmp, "aeon"))
```
psychocrypt: This line is wrong must be `!condition`

xmrstak/jconf.cpp
```diff
@@ -157,14 +158,14 @@ const std::string jconf::GetCurrency()
 	if(currency.empty())
 		currency = prv->configValues[sCurrency]->GetString();
 	if(
-#ifndef CONF_NO_XMR
-			currency.compare("xmr") != 0
+#ifndef CONF_NO_MONERO
+			xmrstak::strcmp_i(currency, "monero")
```
psychocrypt: wrong must be `!condition`

xmrstak/cli/cli-miner.cpp
psychocrypt: well spotted: I inserted two logic bugs during the renaming

- fix with the last commit

fireice-uk:
Ready to merge on my side.

psychocrypt:
frome my side also

PR #68, opened 2017-10-24
amd: use 64bit target

psychocrypt:
- remove 32bit target value
- use always 64bit target for amd backend

# Test

- [x] Tested with OpenCL on NVIDIA
- [x] need test on a AMD GPU *

\* can someone please test this changes


fireice-uk:
Tested on an RX480. Looks ok - merging

PR #69, opened 2017-10-26
thread affinity for non cpu backends

psychocrypt:
This is a follow up of #43 and use the some mechanism to set the thread affinity for
non cpu backends correct.

- use cpu affinity workflow for nvidia and amd
- amd/nvidia: bind host memory with numaloc if affinity is given
- cpu: move messages of thread spawning before thread creation

### DIFF
--- a/xmrstak/backend/amd/minethd.cpp
+++ b/xmrstak/backend/amd/minethd.cpp
@@ -34,6 +34,7 @@
 #include "xmrstak/misc/executor.hpp"
 #include "xmrstak/misc/environment.hpp"
 #include "xmrstak/params.hpp"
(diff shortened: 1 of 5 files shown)


fireice-uk:
@psychocrypt You removed the check for _no affinity_ setting (affinity == -1). Can you add that back in, together with relevant messages?

psychocrypt:
@fireice-uk well spotted, I added the check also to the cpu backend.

psychocrypt:
The messages are placed to the position where they are in the cpu version.
I can move all thraed affinity massages into the thraed to keep it near to
the function.
I will to it tonight.

PR #72, opened 2017-10-27
increase safety memory for autosuggestion

psychocrypt:
- increase safty memory from 64 to 128 MiB
- NVIDIA: increase lmem reserve per thread to 1kiB

This should avoid that the autosuggestion create a config which use to much memory.

### DIFF
--- a/xmrstak/backend/amd/autoAdjust.hpp
+++ b/xmrstak/backend/amd/autoAdjust.hpp
@@ -97,8 +97,8 @@ private:
         int i = 0;
         for(auto& ctx : devVec)
         {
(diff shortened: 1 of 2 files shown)


fireice-uk:
@psychocrypt Is this going to fix crashes experienced by @JerichoJones in #72

psychocrypt:
@fireice-uk  The link to @JerichoJones issue is wrong. But we had a lot of issues in the past where we hit the memory wall during the runtime. 
The reason is in the most cases the dynamic usage of `lmem` (register outsourced to the gpu main memory). Be more conservative with the memory should create more stable configurations.

PR #80, opened 2017-10-29
add option: flush stdout

psychocrypt:
Added the option to forcefully flush STDOUT, which may be useful when running the miner through a pipe instead than a shell.
If STDOUT is a shell, it's unbuffered by default.
If STDOUT is a pipe, it's buffered by default.

This PR is ported from https://github.com/fireice-uk/xmr-stak-cpu/pull/352 to the new miner base.


fireice-uk:
@darugnaa Thanks for contributing!

Issue #88, opened 2017-11-03
Autoconfiguration for GPU way off


sheepchen:
Hey,
I compiled the xmr-stak version for windows and tested it.

My mainworkingstation (Testsystem):
Windows 10 Pro 64 Bit
Intel i7-4790K 4Ghz uc@3.6 Ghz
Nvidia Geforce GTX 980 4GB

configs (generated automaticly):
```
"cpu_threads_conf" :
[
    { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 0 },
    { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 2 },
    { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 4 },
    { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 6 },

],
```
```
"gpu_threads_conf" :
[
  // gpu: GeForce GTX 980 architecture: 52
  //      memory: 3376/4096 MiB
  { "index" : 0,
    "threads" : 20, "blocks" : 48,
    "bfactor" : 6, "bsleep" :  25,
    "affine_to_cpu" : false,
  },

],
```
First run crashed directly:
![grafik](https://user-images.githubusercontent.com/32341444/32369184-9935307e-c088-11e7-953c-a774e041181e.png)

I changed nvidia config like this:
bfactor: 8
bsleep: 100

result: program runs but the System was very laggy.

I changed nvidia config again like this:
bfactor: 12
bsleep: 100

result: program runs and system is only a little bit laggy (an vlc-video runs after 2 sec smootly)

after 15 minutes:
```
HASHRATE REPORT
| ID | 10s |  60s |  15m | ID | 10s |  60s |  15m |
|  0 | (na) | 84.3 | 84.8 |  1 | 31.1 | 31.5 | 32.9 |
|  2 | 38.9 | 37.2 | 37.9 |  3 | 38.2 | 36.3 | 36.1 |
|  4 | 21.8 | 22.0 | 22.9 |
-----------------------------------------------------
Totals:   (na) 211.2 214.5 H/s
Highest:  0.0 H/s
```

So i tweaked the config a little bit:
```
"cpu_threads_conf" : 
[ 
     { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 2 },
     { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 3 },
[...]

PR #89, opened 2017-11-03
optimize NVIDIA autosuggestion

psychocrypt:
- avoid creation of a config with zero threads or blocks
- WINDOWS: reduce the used memory for the auto suggestion by the amount of already used memory

We use in windows bfactor (split slow kernel into smaller parts) to avoid that windows is killing long running kernel.
In the case there is already memory used on the gpu than we assume that other application are running between the split kernel, this can result into TLB memory flushes and can strongly reduce the performance and the result can be that windows is killing the miner.
Be reducing maximum memory for the autosuggestion be the amount of the current used memory we try to avoid this effect.

- [x] do not merge before we get feedback from @sheepchen

This PR is related to #88

### DIFF
--- a/xmrstak/backend/nvidia/autoAdjust.hpp
+++ b/xmrstak/backend/nvidia/autoAdjust.hpp
@@ -88,17 +88,18 @@ private:
 
 		constexpr size_t byte2mib = 1024u * 1024u;
 		std::string conf;
-        int i = 0;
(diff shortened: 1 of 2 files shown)


psychocrypt:
This can be merged. PR is working as expected https://github.com/fireice-uk/xmr-stak/issues/88#issuecomment-341861816

PR #96, opened 2017-11-07
Install libraries to the library directory

candrews:
Libraries should be installed to "lib" not "bin"

For example, `libxmrstak_opencl_backend.so` is a library to it should be installed to "lib" not "bin"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -456,7 +456,7 @@ add_executable(xmr-stak
 )
 
 set(EXECUTABLE_OUTPUT_PATH "bin")
-set(LIBRARY_OUTPUT_PATH "bin")
+set(LIBRARY_OUTPUT_PATH "lib")
 
 target_link_libraries(xmr-stak ${LIBS} xmr-stak-c xmr-stak-backend)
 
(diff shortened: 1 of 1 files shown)


fireice-uk:
@psychocrypt Can you confirm if this change won't mess with appimage work?

psychocrypt:
Thx for opening this pull request. The libraries are in bin to keep the
setup easy.
If I move the libraries (plugins in our case) to an other folder than we
need to set in linux e.g. LD_LIBRARY_PATH. this is to complicated for an
non advanced user.
Maybe we can add an option in cmake that the user can change this behavior
this will simplify the packing for linux distributations.
IMO there is no need to have it in the upcoming release.

psychocrypt:
@fireice-uk I move the libs in appImage into a seperate lib folder.
Never the less I can add a option after the release but before. The user
friendly way is to keep it together.

candrews:
This is user friendly, and it's how pretty much everything else works, and doing so is a definitely requirement for being packaged by any linux distro (I'm a Gentoo developer currently working on getting xmr-stak included in Gentoo).

No code changes are required - users would not be required to set `LD_LIBRARY_PATH`. I just tested this by simply moving /usr/bin/libxmrstak_opencl_backend.so to /usr/lib and xmr-stak still worked fine (including detecting my amd opencl device). The reason is that `dlopen` checks `lib` and `LD_LIBRARY_PATH` (if set), see https://linux.die.net/man/3/dlopen Which also means that plugin.hpp could and should be greatly simplified.

I'm not sure what use case you're trying to satisfy by having the executable binary and the libraries in the same directory, as AppImage doesn't require that (it includes support for  `lib` directory): https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages

psychocrypt:
Yes it is working in your case because you installed the miner to the root
file structure. Not all other have root access and moat of it start the
miner directly from the build folder.
I will add soon option to make the packaging process easier. Currentlu we
are also lacking suppory to have a environment variable to sau where the
config files should be stored.
By default all config files should be located in $HOME/.config/xmr-stak and
not to the current working directory.

fireice-uk:
@psychocrypt I would much rather keep everything in install directory by default, I don't consider a miner to be your usual run-of-the-mill system tool. Keeping everything in one place makes it easier for non-technical users to see what's going on.

psychocrypt:
I will close this PR:
I will add packaging options after our release. I added #118 for this job

PR #114, opened 2017-11-14
fix wrong memory detection

psychocrypt:
Free and total memory is only evaluated on the first device.
To detect the gpu memory the gpu must be selected.

- create context on the gpu before the memory is checked
- add smx to the auto detection
- change the result code of `cuda_get_deviceinfo()`

This bug is also included in the xmr-stak-nvidia repository.

### DIFF
--- a/xmrstak/backend/nvidia/autoAdjust.hpp
+++ b/xmrstak/backend/nvidia/autoAdjust.hpp
@@ -60,17 +60,15 @@ public:
             ctx.device_bfactor = 6;
             ctx.device_bsleep = 25;
 #endif
-            if( cuda_get_deviceinfo(&ctx) != 1 )
-            {
-                printer::inst()->print_msg(L0, "Setup failed for GPU %d. Exitting.\n", i);
-                std::exit(0);
-            }
-            nvidCtxVec.push_back(ctx);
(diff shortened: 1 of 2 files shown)


xmrstak/backend/nvidia/autoAdjust.hpp
```diff
@@ -60,12 +60,10 @@ class autoAdjust
             ctx.device_bfactor = 6;
             ctx.device_bsleep = 25;
 #endif
-            if( cuda_get_deviceinfo(&ctx) != 1 )
-            {
-                printer::inst()->print_msg(L0, "Setup failed for GPU %d. Exitting.\n", i);
-                std::exit(0);
-            }
-            nvidCtxVec.push_back(ctx);
+            if(cuda_get_deviceinfo(&ctx) == 0)
```
fireice-uk: Can you fix the space indents here, then cherrry-pick 96fa0cec onto a fresh branch (fixing this will cause merge conflict with #116).

psychocrypt:
Ahh right i will change it to spaces and if all our PR are merged I will
fix at globaly.

psychocrypt:
I changed all to space indention and rebased the branch which depend on this PR

PR #115, opened 2017-11-14
Network extensions

fireice-uk:
This is my final feature PR before the release. It implements algo, hashcount, backend and motd extensions.

With regards to motd extension:
- Only English characters supported for now (Unicode needs more thought from the security angle)
- Javascript injection is considered and prevented
- Terminal characters and other rubbish is filtered.

Needs testing, but should be ok. I will update the mock pool to include extensions.

### DIFF
--- a/xmrstak/backend/amd/minethd.cpp
+++ b/xmrstak/backend/amd/minethd.cpp
@@ -237,7 +237,7 @@ void minethd::work_main()
 
 				hash_fun(bWorkBlob, oWork.iWorkSize, bResult, cpu_ctx);
 				if ( (*((uint64_t*)(bResult + 24))) < oWork.iTarget)
-					executor::inst()->push_event(ex_event(job_result(oWork.sJobID, results[i], bResult), oWork.iPoolId));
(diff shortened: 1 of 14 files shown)


fireice-uk:
Last thing. I changed the backend string type to a static array since it will be invoked with every result.

fireice-uk:
I pushed an update to the mock pool that adds extensions to enable testing ( https://github.com/fireice-uk/xmr-pool-mock ) -> please test weird and wonderful ways of breaking it (HTML, binary data, random garbage etc).

The correct operation of other extensions can be monitored on the output dump.

psychocrypt:
I played a little bit around with the new feature.

# feedback 
- I am confused by the name `motd` I started the miner and got no message from the pool during the first connection. Than I checked the code and realized that the message is shown with the hash report.
- If I have to pools in my list I get the motd from all pools, not only from the where I am connected to.
- If I have two pools and one is going offline I still get the `motd` shown.

fireice-uk:
1. motd = "Message of the day"
2. This is by design. I assume if you have a pool in your pool list you will be interested in downtimes, pool parties etc.
3. Again by design. The miner should not connect simply to fetch the message so it will store it when not connected.

psychocrypt:
thx, thus my testings yesterday shows no issues I will merge it

PR #134, opened 2017-11-17
Fix GPU result crash

fireice-uk:
Fairly trivial issue - forgot to add copy code so the application tried to read random memory.

I added an explicit constructor too, just to make it clearer what is happening.


fireice-uk:
Will merge it as soon as CI passes.

psychocrypt:
Is this fix related to an open issue?

fireice-uk:
Yes, xref'ed it, refresh the page if you missed it :)

PR #136, opened 2017-11-18
fix OpenCL compatibility

psychocrypt:
- enforce that only host side OpenCl 1.2 functions are used if the miner is compiled with `XMR-STAK_COMPILE=generic`
- update documentation
- add CMake option `XMR-STAK_COMPILE` to windows

On some linux system the provided OpenCL header will set OpenCl 2.0 available but the library (e.g. Mesa) is only shipping OpenCl 1.2. This will crash the miner.
This is also important to ship pre-compiled binaries with a high system compatibility (Linux and Windows)


fireice-uk:
@psychocrypt I tend to build the release miners against newest libs from nvidia and amd (9.0 and 3.0 respectively), is that ok?

psychocrypt:
@fireice-uk Please do not  use Cuda9. This will avoid the usage of fermi GPUs. 
A NVIDIA C2070 give 150H/s for XMR and over 350 H/s for Aeon. 
AMD AppSDK should be fine if `gerenic` is enabled.

BTW: You can install the NVIDIA Toolkit 8.0 and 9.0 under windows at the same time and call `cmake -D CUDA_TOOLKIT_ROOT_DIR=<path to cuda8> ..`.

Issue #129, opened 2017-11-17
One user's problem


fireice-uk:
One guy on IRC managed to get 12 (!) gpus working with xmr-stak-amd, but they aren't detected by xmr-stak - some pastebins:

https://thepasteb.in/p/3lhJcDkOW4D18hN (needs sudo to actually see the gpus)

https://thepasteb.in/p/P1hrcj5mx7l3vcX (with a working config from stak-amd)

https://thepasteb.in/p/vghjcAxwqkKxxhx (I guided him into adding the line dumping number of devices and platform)

@psychocrypt -  can you think of anything?

psychocrypt:
Is the user using the AMD blockchain driver?
If so is it the latest beta version.
There are a lot of issues with the blockchain driver in the past. The
issues are equal to this one. The OpenCL call `clGetDeviceIDs` is not
giving the correct number of devices or fails.

Is it possible to update to the newest or to an older driver?

PR #137, opened 2017-11-18
handle OpenCL errors

psychocrypt:
- handle opencl errors to avoid unexpected behaviors
- remove code duplicatio nwithin the some method

Handle the most important OpenCL API calls to avoid that e.g. the device list is created if the API call to get the number of devices is failing.
This can maybe help to analyse #129

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/gpu.cpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.cpp
@@ -432,8 +432,11 @@ uint32_t getNumPlatforms()
 
 	// Get platform and device information
 	clStatus = clGetPlatformIDs(0, NULL, &num_platforms);
-	platforms = (cl_platform_id *) malloc(sizeof(cl_platform_id) * num_platforms);
-	clStatus = clGetPlatformIDs(num_platforms, platforms, NULL);
+	if(clStatus != CL_SUCCESS)
+	{
+		printer::inst()->print_msg(L1,"WARNING: %s when calling clGetPlatformIDs for number of platforms.", err_to_str(clStatus));
+		return 0u;
+	}
 
 	return num_platforms;
 }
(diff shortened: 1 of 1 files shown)


xmrstak/backend/amd/amd_gpu/gpu.cpp
```diff
@@ -448,39 +451,60 @@ std::vector<GpuContext> getAMDDevices(int index)
...
-	clStatus = clGetDeviceIDs( platforms[index], CL_DEVICE_TYPE_GPU, 0, NULL, &num_devices);
-	device_list = (cl_device_id *) malloc(sizeof(cl_device_id)*num_devices);
-	clStatus = clGetDeviceIDs( platforms[index], CL_DEVICE_TYPE_GPU, num_devices, device_list, NULL);
-	for (int k = 0; k < num_devices; k++) {
-		cl_int clError;
-		std::vector<char> devVendorVec(1024);
-		clError = clGetDeviceInfo(device_list[k], CL_DEVICE_VENDOR, devVendorVec.size(), devVendorVec.data(), NULL);
-		std::string devVendor(devVendorVec.data());
-		if( devVendor.find("Advanced Micro Devices") != std::string::npos)
+	if(numPlatforms)
```
fireice-uk: Accepting right now, but this will need to be reformatted.
Good way to do it is:
```
std::vector<GpuContext> getAMDDevices(int index) -> bool getAMDDevices(int index, std::vector<GpuContext>& ctxVec)

if(!numPlatforms)
    return false;
```

psychocrypt: You are right I wrote it on this way to avoid that I need to call `free` to the created heap buffer in the case of an error/warning.
The full files needs a cleanup. There is a lot of dead code in and many code duplication's which should be removed to increase the maintenance.

fireice-uk: This is actually a good usecase for a dynamic structure, you can avoid it by doing `std::vector<uint_8t> buffer; buffer.resize(len); buffer.data()`

PR #143, opened 2017-11-19
Add windows exit

fireice-uk:
This should fix the version issue.


fireice-uk:
Win CI finished, but locked up (win_exit lol)... We will need to fix that later.

PR #163, opened 2017-11-20
fix undefined std::string

ExceptionallyGreat:
Fixing MacOS X Compile error

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/gpu.hpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.hpp
@@ -8,10 +8,8 @@
 #include <CL/cl.h>
 #endif
 
-#include <string>
-#include <sstream>
-#include <iostream>
 #include <stdint.h>
+#include <string>
 #include <vector>
 
 #define ERR_SUCCESS (0)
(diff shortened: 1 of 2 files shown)


xmrstak/backend/amd/amd_gpu/gpu.hpp
```diff
@@ -8,6 +8,9 @@
 #include <CL/cl.h>
 #endif
 
+#include <string>
+#include <sstream>
```
grzegorzszczecin: Notice that just including <string> seems to be sufficient; the <sstream> and <iostream> are not needed to fix the compile error. Also please put this below <stdint.h> so that heades are included in alphabetic order

ExceptionallyGreat:
Updated as per suggestion.

grzegorzszczecin:
thanks this fixes the compile error on my Mac

psychocrypt:
THX for this PR!
#169 contains the some information for the macOSX installation therefore we will merge it after it is opened against the dev branch.
Could you please remove the second commit, that we can merge the compile fix only.

note: please do not merge different topics within one PR. The first commit is a bug fix and we can label it with the BUG label. The second is a documentation improvement.

psychocrypt:
please rebase you PR against the dev branch after #179 is merged. This will fix the windows CI tests with appvoyer.

ExceptionallyGreat:
Deleted the compile_macos_x.md file. I am new to github, so I am not sure if that would work.

PR #179, opened 2017-11-20
fix appvoyer

psychocrypt:
Remove the execution of the binary at the end of the CI test. This is only currently a temporary solution.

The windows CI is broken since #143


fireice-uk:
@grzegorzszczecin On Windows builds we wait for "any key" before closing to enable most users to simply double click the binary. On appveyor the input never comes, so the build times out.

PR #181, opened 2017-11-20
update windows compile guide  

### LINKED ISSUE #153: instructions to compile on windows do not work
Greetings!

I have followed instructions to compile on windows 10.  I did not perform steps for CUDA 8.0+ nor did I download and create the directory structure for the dependency folder - again, no CUDA needed; I have a Vega.  When I create the build directory and cd to it and then run the cmake command, I get `Error: could not load cache`

Why am I changing to an empty directory and running cmake without specifying the location of make files or whatnot?  Anyway, I would like this to work, please.  :)  Please look at your instructions and let me know how I screwed up.  Thanks!

psychocrypt:
Add explicit note that VSC v141 should be used for AMD and CPU compile only.

fix #153 

- [x] please merge after #179. Appvoyer fix is included in the first commit of the PR

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -29,4 +29,4 @@ test_script:
   - cd c:\xmr-stak\build\bin\Release
   - dir
   - copy C:\xmr-stak-dep\openssl\bin\* .
-  - xmr-stak.exe --help
+#  - xmr-stak.exe --help
--- a/doc/compile_Windows.md
+++ b/doc/compile_Windows.md
@@ -81,15 +81,19 @@
   mkdir build
   cd build
   ```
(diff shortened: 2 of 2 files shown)


saylor68:
dunno if this is proper spot to comment - github newb here - my note is that the config file needs to be in same working directory as exe so it can find the config file when launching the exe. 1st time I ran this it said it couldn't find the config file..
thanks!
- saylor

psychocrypt:
@fireice-uk  This can be merged now

psychocrypt:
@saylor68 Thx for you feedback. I will think about where the best place is to add this behavior to documentation.

PR #188, opened 2017-11-21
clarified a third way to change variables

Clintoxen:
Added an easier way for inexperienced people to change the build options variables


psychocrypt:
Thanks for you suggestion.
Sry that I need to say that we will not accept this PR.
The reasons is if the user is changing something within the `CMakeLists.txt`  and than create a syntax error or similar it is for us impossible to give support. 
All options are exposed to the user and can be changed with `-DVARNAME=VALUE`.

Clintoxen:
Alright I understand

PR #193, opened 2017-11-21
Fix uninitialized memory accces

blacklion:
Uninitialized memory access is UB in C/C++, fix it.


fireice-uk:
:+1: one note -> faster and cleaner way to write this in C++ is ` = {0}`

PR #186, opened 2017-11-21
fixed undefined std::string

ChromeHearts:
Tested on MAC OSX Sierra

PR #194, opened 2017-11-21
Update gpu.hpp

ExceptionallyGreat:
Fixing the compile error on MacOS X.
Close the old PR because of issues with GitHub. Opening a new one.

PR #195, opened 2017-11-21
fixed undefined std::string

psychocrypt:
Tested on MAC OSX Sierra

This PR is equal to #186 but opened against the dev branch.


psychocrypt:
already merged with #194. The CI was faster with the othe rpull request.
@ChromeHearts Thanks for your PR

PR #215, opened 2017-11-23
Adding Dockerfile and fixing build_xmr-stak_docker script

ruzickap:
Adding Dockerfile and fixing build_xmr-stak_docker script.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,39 @@
+# Latest version of ubuntu
+FROM ubuntu
+
+# Default git repository
+ENV GIT_REPOSITORY https://github.com/fireice-uk/xmr-stak.git
+ENV CUDA_URL https://developer.nvidia.com/compute/cuda/8.0/prod/local_installers/cuda_8.0.44_linux-run
+ENV OPENCL_URL http://pages.cs.wisc.edu/~riccardo/assets/AMD-APP-SDKInstaller-v3.0.130.136-GA-linux64.tar.bz2
+
+ENV CMAKE_C_COMPILER gcc-5
+ENV CMAKE_CXX_COMPILER g++-5
+ENV XMRSTAK_CMAKE_FLAGS -DXMR-STAK_COMPILE=generic -DCUDA_ENABLE=ON -DOpenCL_ENABLE=OFF
+
+# Innstall packages
+RUN apt-get update \
+    && set -x \
+    && apt-get install -qq --no-install-recommends -y ca-certificates cmake g++ gcc git libhwloc-dev libmicrohttpd-dev libssl-dev lsb-release wget \
+    && wget -q $CUDA_URL \
+    && chmod u+x /cuda_*_linux-run \
+    && /cuda_*_linux-run --silent --toolkit \
+    && wget -q $OPENCL_URL -O - | tar xjf - -C / \
+    && /AMD-APP-SDK-*-linux64.sh --keep -- --silent --acceptEULA=yes \
+    && . /etc/profile.d/AMDAPPSDK.sh \
+    && git clone $GIT_REPOSITORY \
+    && cd /xmr-stak \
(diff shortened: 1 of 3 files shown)


scripts/build_xmr-stak_docker/build_xmr-stak_docker.sh
```diff
@@ -5,110 +5,116 @@ if [[ $EUID -ne 0 ]]; then
...
 ########################
-docker run --rm -it -v $PWD/xmr-stak-cpu:/xmr-stak-cpu fedora:26 /bin/bash -c "
+docker run --rm -it -v $PWD/xmr-stak:/xmr-stak fedora:27 /bin/bash -c "
 set -ex ;
 dnf install -y -q gcc gcc-c++ hwloc-devel libmicrohttpd-devel libstdc++-static make openssl-devel cmake ;
-cd /xmr-stak-cpu ;
-cmake -DCMAKE_LINK_STATIC=ON . ;
-make install ;
+cd /xmr-stak ;
+cmake -DCUDA_ENABLE=OFF -DOpenCL_ENABLE=OFF . ;
```
psychocrypt: why do you build without CUDA and OpenCL support?

CUDA and OpenCL  install is for all linux something like the following:
```
ENV PATH=/CUDA_ROOT=/usr/local/cuda-8.0

RUN  wget https://developer.nvidia.com/compute/cuda/8.0/prod/local_installers/cuda_8.0.44_linux-run && \
    chmod 755 ./cuda_8.0.44_linux-run && \
    ./cuda_8.0.44_linux-run --silent --toolkit --toolkitpath=$CUDA_ROOT && \
    rm ./cuda_8.0.44_linux-run
```

psychocrypt:
Thanks for updating the docker files. Could you please enable CUDA and OpenCL.

We can also switch to CUDA9 and use the nvidia provided docker image. It is since CUDA9 also as small package available.

https://github.com/NVIDIA/nvidia-docker/wiki/CUDA
```
FROM       nvidia/cuda:9.0-base
```

scripts/build_xmr-stak_docker/build_xmr-stak_docker.sh
ruzickap: I added CUDA wherever it was supported by the installer.

ruzickap:
The reason why I'm not using the suggested image ([nvidia/cuda](https://hub.docker.com/r/nvidia/cuda/)) is, that image has already CUDA libraries inside and it's quite big (1 GB).

The image I'm using is minimal ubuntu and build process will install everything what is needed, build the xmr-stak binaries and then remove the the compilers/libraries/src codes which are not used - so the final image is not so big (451 MB) and contains only xmr-stack + nvidia library.

psychocrypt:
note: since cuda9 the docker of images have different layers and the size
is much smaller if only the driver and the toolkit is needed.

ruzickap:
Ok ...
I changed the Dockerfile to use the "nvidia/cuda:9.0-base" and it's not so big as I expected.
Please check it and le me know.
Thank you.

Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Latest version of ubuntu
+FROM nvidia/cuda:9.0-base
+
+# Default git repository
+ENV GIT_REPOSITORY https://github.com/fireice-uk/xmr-stak.git
+ENV XMRSTAK_CMAKE_FLAGS -DXMR-STAK_COMPILE=generic -DCUDA_ENABLE=ON -DOpenCL_ENABLE=OFF
```
psychocrypt: Note: We can enable OpenCL because it is shipped with CUDA. Maybe not with the base image. But we can do it in a follow up PR.

psychocrypt:
Thanks again for updating the docker files +1

ruzickap:
By the way - @fireice-uk may I ask you to create [automated Docker build](https://docs.docker.com/docker-hub/builds/#create-an-automated-build) in Docker Hub for xmr-stak please?

PR #269, opened 2017-11-27
fixed releases link in documentation

jaymedavis:


psychocrypt:
Thanks for you pull request. The issue is already fixed with #259.

Note: please open pull request always against the dev branch. The master is only updated if we release a new version. https://github.com/fireice-uk/xmr-stak/blob/79154f76defa627255891888f4dd62453194fc2c/CONTRIBUTING.md

PR #279, opened 2017-11-28
Add macOS CPU-only build instructions

jamwaffles:
The process is pretty simple, but it's worth noting down for future
visitors to the project. I tested it on a 2017 Macbook Pro. I didn't
include OpenCL/AMD instructions as I have no way of testing them.
Whoever wants to start mining on a new iMac Pro can add those instead ;)


jamwaffles:
Never mind! Rebasing on `dev` showed there are already better docs for macOS systems.

PR #305, opened 2017-11-30
allow to diable UAC dialog

psychocrypt:
- remove CMake option `WIN_UAC`
- spawn UAC dialog via restarting xmr-miner with administrator right
- allow to disable UAC with `--noUAC`
- update documentation
- remove usage section with help message (output depends on OS)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,7 +22,7 @@ build_script:
   - mkdir build
   - cd build
   - set CMAKE_PREFIX_PATH=C:\xmr-stak-dep\hwloc;C:\xmr-stak-dep\libmicrohttpd;C:\xmr-stak-dep\openssl;
-  - cmake -G "Visual Studio 14 2015 Win64" -T v140,host=x64 .. -DWIN_UAC=OFF -DCUDA_ARCH=30
+  - cmake -G "Visual Studio 14 2015 Win64" -T v140,host=x64 .. -DCUDA_ARCH=30
   - cmake --build . --config Release --target install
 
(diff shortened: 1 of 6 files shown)


fireice-uk:
I should get some time to test it this Sunday.

fireice-uk:
@psychocrypt This PR spawns infinite processes when launched on Win 7 with a double-click

psychocrypt:
Arrggs I have currently no windows7. I think the option --noUAC is not
detected after the start of the process with high privilegs.
Could you debug it? Else I will do is in the next two days.

fireice-uk:
Way too busy today. I just managed to test this and #313 , I will merge that one when this is fixed (my Nvidia system runs on Win7 while AMD on Linux).

psychocrypt:
@fireice-uk I checked this code on two independent windows 7 systems and can not reproduce your infinite loop. Could you please test it with a clean code version again. 

[This](https://github.com/fireice-uk/xmr-stak/pull/305/files#diff-6c13e8fdffaf680d2e2876486b6124abR286) line should avoid that the miner starts in an infinite loop. If you still get this error could you please print `argv` in the first line of the main.

psychocrypt:
ready to merge

Simon0Harms:
> On Windows 10 it is only needed once to set up the account to use them.

Can you explain this?

fireice-uk:
On Windows the user needs to have a special permission to use large pages. The miner will auto-magically do that if the user hasn't got them, but it needs to be ran as administrator.

PR #313, opened 2017-12-01
option to controll gpu synchronization

psychocrypt:
- add option `sync_mode`
- update auto suggestion and jconf

Setting `sync_mode` to `2` can increase the gpu hash rate by up to `4%` but will create some load on one thread per gpu. The default will be the same then before.


fireice-uk:
Just noting here that in my tests all options except the default decreased performace

PR #315, opened 2017-12-01
add AMD scratchpad indexing option

psychocrypt:
Allow to change the indexing used to address the hash scratchpad memory.

- add option `strided_index` for each gpu

I changed the indexing in the old amd version https://github.com/fireice-uk/xmr-stak-amd/pull/35. Some users reported a hash rate decrease. This option will give the user the opportunity to tune the indexing depending of the hardware. ~~The default is equal to the current dev.~~ 
The default option will be `true` which gives a few hashes more on newer hardware. I changes it in the past based on my results of a HD7870.

- [x] tested on NVIDIA gpu's with OpenCL
- [x] tested and benchmarked from the [reddit community]((https://www.reddit.com/r/MoneroMining/comments/7gygtb/amd_xmrstak_help_wanted/?ref=share&ref_source=link) - thx @all tester for the feedback

### DIFF
--- a/doc/tuning.md
+++ b/doc/tuning.md
@@ -8,6 +8,7 @@
   * [Choose `intensity` and `worksize`](#choose-intensity-and-worksize)
   * [Add more GPUs](#add-more-gpus)
   * [Increase Memory Pool](#increase-memory-pool)
+  * [Scratchpad Indexing](#scratchpad-indexing)
 
 ## NVIDIA Backend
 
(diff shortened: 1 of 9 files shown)


psychocrypt:
@anhphan Could you please test if the new option has a positive effect for vega gpus. You need to create a new auto suggested `amd.txt` to get the new option passed into the config.

psychocrypt:
I forced push a fix for wrong index calculation.

psychocrypt:
Based on the community feedback I set `strided_index` to `true` as default.

fireice-uk:
Great. Merging together with #332 for combined testing.

1RedOne:
This option is listed on the Readme, but it doesn't go into depth as to why or when someone would want to tweak this setting.  Could a blurb be added to that affect?  We see lots of forum threads due to vagueness here.

PR #345, opened 2017-12-04
[docs] Explicitly linked to an implied location

TimothyShimmin:
I expect that link is something that would be easily lost if another item is added.


fireice-uk:
Thanks!

PR #90, opened 2017-11-04
Multi-pool final version

fireice-uk:
This is a very first draft. Lots of deep-running logic changes, so bugs are inevitable.

Uploading this one early so it can be tested. Dev donate is 50% / 4min in this branch to facilitate testing.

Only pool failover is supported. We can leave round-robin until after the release as this is likely to be even more complex.

Config side is not implemented at all yet, you can add pools directly in executor main.

I implemented a mock pool to facilitate testing, see http://github.com/fireice-uk/xmr-pool-mock

**Final tasks before merging**

- [x] Set a different dev pool port for Aeon
- [x] Restore dev pool address
- [x] Restore dev doation to normal level

PR #378, opened 2017-12-06
fix broken pool option overwrite via CLI

### LINKED ISSUE #340: Config file override by command line possible?
Hello.

Running under Ubuntu 16.04 LTS. This miner is the first one I've come across that cannot run w/o a config.txt file.  Other miners use config files too, but they are not mandatory. I usually add rig name after wallet address like $WALLET.$RIG with RIG set to (hostname). This parameter is passed via command line and it works with other miners.

But XMRSTAK picks up parameters from the config.txt once it has been created and it **ignores** newer params passed via command line (contrary to what -h command claims). I could erase the config.txt on every run, however, XMRSTAK enters a kind of interactive state when it asks questions about coin name and pool parameters. 

So, if I had, say, 100 rigs, I would have to run XMRSTAK a 100 times to create an individual config file for every rig.

p.s. I just realized your group name is FIRE ICE-UK. If written as one word, it reads "fierce-uk".

psychocrypt:
fix #340

With #90 the option to overwrite the pool settings via CLI options is broken.

This pull request will allow to overwrite the options of the pool with the highest weighting.

### DIFF
--- a/xmrstak/cli/cli-miner.cpp
+++ b/xmrstak/cli/cli-miner.cpp
@@ -82,7 +82,7 @@ void help()
 	cout<<"  --nvidia FILE         NVIDIA backend miner config file"<<endl;
 #endif
 	cout<<" "<<endl;
-	cout<<"The Following options temporary overwrites the config file settings:"<<endl;
+	cout<<"The Following options temporary overwrites the config entries of \nthe pool with the highest weight:"<<endl;
 	cout<<"  -o, --url URL         pool url and port, e.g. pool.usxmrpool.com:3333"<<endl;
 	cout<<"  -u, --user USERNAME   pool user name or wallet address"<<endl;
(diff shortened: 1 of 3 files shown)


xmrstak/cli/cli-miner.cpp
```diff
@@ -82,7 +82,7 @@ void help()
 	cout<<"  --nvidia FILE         NVIDIA backend miner config file"<<endl;
 #endif
 	cout<<" "<<endl;
-	cout<<"The Following options temporary overwrites the config file settings:"<<endl;
+	cout<<"The Following options temporary overwrites the config entries of \nthe pool with the highest weighting:"<<endl;
```
fireice-uk: Sp. `weight` not `weighting`

psychocrypt:
@fireice-uk done

PR #380, opened 2017-12-07
CPU ONLY Docker alpine build

rowanruseler:
Created an alpine build for CPU only farming, always good to support xmr
with some cpu compute of available or lingering devices.

Please see README.md for more information

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -29,4 +29,6 @@ VOLUME /mnt
 WORKDIR /mnt
 
 COPY ["run.sh", "setup.sh", "./"]
+
+RUN chmod +x run.sh
 CMD ["./run.sh"]
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -6,15 +6,15 @@ RUN apk --no-cache upgrade \
                 automake \
                 autoconf \
                 openssl-dev \
-                build-base \
                 git \
+                build-base \
+                libmicrohttpd-dev \
         && git clone https://github.com/fireice-uk/xmr-stak.git \
         && cd xmr-stak \
         && cmake . \
                 -DCUDA_ENABLE=OFF \
                 -DOpenCL_ENABLE=OFF \
                 -DHWLOC_ENABLE=OFF \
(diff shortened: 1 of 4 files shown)


lucaforni:
Hi,
I tried to run your docker but when i have this error where the container start:
docker: Error response from daemon: oci runtime error: container_linux.go:265: starting container process caused "exec: \"./run.sh\": permission denied".

Docker version 17.09.0-ce, build afdb6d4 on OSX 10.13.1
Any idea?
Thank you

rowanruseler:
@lucaforni I decided not to grand execute rights (chmod +x) in the Dockerfile, but instead just to grand it in the repo myself.

Can you attempt to give it execute rights and try again?

lucaforni:
Ok,
It works
Thank you

rowanruseler:
@lucaforni just added it to the Dockerfile, nice catch. It should be easier for people to run if it is just another layer within the Dockerfile.

If you have any other issues, let me know! Cheers.

lucaforni:
There is a type in setup.sh:
POOL_ADDRESS=${POOL_ADDRESS:-pool.supportxmr.com:3333}

psychocrypt:
Thank you for your pull request. 
 
I am not sure if the way how you created the config file is a good way. The problem is that we need to check you scripts each time we add now options to the config file e.g. #315 and #313.
I think it is not possible for use to maintain the scripts.

Is it possible that you create a own repository with your extensions. We can later on link it somewhere in the documentation (in a section with useful extensions). I think a own project with xmr-stak as submodule would be good. This will give you the possibility to point to a exact version and a update from our side will not break the scripts.

rowanruseler:
@psychocrypt I would love to maintain it, so I did as adviced: https://github.com/rowanruseler/xmr-stak-cpu-alpine

Let me know if you want some addional credits or anything else in it, cheers.

krll-k:
@rowanruseler can I skip the video cards through the docker?

rowanruseler:
@krll-k 'skip' as in "not use them" or 'skip' as in "run them through docker" ?

PR #391, opened 2017-12-07
Fix cli questions

fireice-uk:
- Fix CLI startup so that it is possible to supply all arguments for instant launch from the command line
- Make the CLI pool a top priority instead of overwriting other pools
- If the CLI pool is missing login / password, but is in the config, they will be supplied from the config

PR has been tested using our mock pool

### DIFF
--- a/xmrstak/cli/cli-miner.cpp
+++ b/xmrstak/cli/cli-miner.cpp
@@ -85,7 +85,8 @@ void help()
 	cout<<"  --nvidia FILE         NVIDIA backend miner config file"<<endl;
 #endif
 	cout<<" "<<endl;
-	cout<<"The Following options temporary overwrites the config entries of \nthe pool with the highest weight:"<<endl;
+	cout<<"The following options can be used for automatic start without a guided config,"<<endl;
+	cout<<"If config exists then this pool will be top priority."<<endl;
 	cout<<"  -o, --url URL         pool url and port, e.g. pool.usxmrpool.com:3333"<<endl;
 	cout<<"  -O, --tls-url URL     TLS pool url and port, e.g. pool.usxmrpool.com:10443"<<endl;
 	cout<<"  -u, --user USERNAME   pool user name or wallet address"<<endl;
@@ -172,7 +173,7 @@ inline void prompt_once(bool& prompted)
 	}
 }
 
-void do_guided_config(bool userSetPasswd)
+void do_guided_config()
 {
 	using namespace xmrstak;
 
@@ -229,7 +230,7 @@ void do_guided_config(bool userSetPasswd)
 	}
 
 	auto& passwd = params::inst().poolPasswd;
-	if(passwd.empty() && (!userSetPasswd))
+	if(passwd.empty() && !params::inst().userSetPwd)
 	{
 		prompt_once(prompted);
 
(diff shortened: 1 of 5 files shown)


psychocrypt:
could you please also expose the nicehash option?

xmrstak/misc/executor.cpp
```diff
@@ -509,49 +511,43 @@ void executor::ex_main()
 			win_exit();
 		}
 #endif
-		if(!cfg.tls) tls = false;
-		pools.emplace_back(i+1, cfg.sPoolAddr, cfg.sWalletAddr, cfg.sPasswd, cfg.weight, false, cfg.tls, cfg.tls_fingerprint, cfg.nicehash);
+		if(!cfg.tls) dev_tls = false;
+
+		if(!xmrstak::params::inst().poolURL.empty() && xmrstak::params::inst().poolURL == cfg.sPoolAddr)
```
psychocrypt: Does this mean that the pass word and username passed by the user via cli params is only used if the user also provide a pool address?

fireice-uk: Yes. What's the point of sending off a username and password to a random pool?

psychocrypt: The point is that a miner operator can add a pool to the config and than overwrite only the password which is in many pools the unique identifier for the miner rig.
The PR implementation has a side effect that the user can not see that the password provided via cli is not used in the case where no pool url is provided via a second cli option.

psychocrypt: If this should be the behavior than we need to add a section somewhere to the documentation and best to the help message.

fireice-uk: My instinct is to veto that on grounds of security issues where login/pass is actually a login pass. I will add an error instead.

xmrstak/cli/cli-miner.cpp
```diff
@@ -440,6 +440,13 @@ int main(int argc, char *argv[])
 		}
 		else if(opName.compare("-u") == 0 || opName.compare("--user") == 0)
 		{
+			if(params::inst().poolURL.empty())
```
psychocrypt: This is not working because the order is not guaranteed.

This is creates an error: `./xmr-stak -c x -u x --url xmr.crypto-pool.fr:333`
This will not create an error `./xmr-stak -c x --url xmr.crypto-pool.fr:333 -u w`

This and the next check must be moved to [this](https://github.com/fireice-uk/xmr-stak/pull/391/files#diff-e6399cdf5a15d37895e060b979c01e31R516) function

xmrstak/cli/cli-miner.cpp
```diff
@@ -451,6 +458,13 @@ int main(int argc, char *argv[])
 		}
 		else if(opName.compare("-p") == 0 || opName.compare("--pass") == 0)
 		{
+			if(params::inst().poolURL.empty())
```
psychocrypt: same like above.

fireice-uk:
I rebased to resolve conflicts with other prs

psychocrypt:
if a pool is used which is not in the json pool list and no password and username is provided than the miner trying to connect but should be abort.

```
./xmr-stak -c x --url xmr.crypto-pool.fr:7777
...
[2017-12-08 21:06:40] : Pool xmr.crypto-pool.fr:7777 connected. Logging in...
[2017-12-08 21:06:40] : SOCKET ERROR - [xmr.crypto-pool.fr:7777] missing login
[2017-12-08 21:06:40] : Pool xmr.crypto-pool.fr:3333 connected. Logging in...
[2017-12-08 21:06:40] : Difficulty changed. Now: 12000.
[2017-12-08 21:06:40] : Pool logged in.
```

PR #396, opened 2017-12-07
Grammar and wording regarding virus protection software

b-:
I chuckled reading that my virus protection software might flag the miner binary as “male ware,” and rewrote that section of the documentation.


fireice-uk:
=) Thanks!

PR #445, opened 2017-12-09
Fix compat_usleep for WIN32

awhittle3:
When compiled with VS2017, the negative applied to the uint wait time is ignored. Fixed by casting first.

This has vastly improved stability on my Windows 10 setup and has increased performance significantly.

### DIFF
--- a/xmrstak/backend/nvidia/nvcc_code/cuda_core.cu
+++ b/xmrstak/backend/nvidia/nvcc_code/cuda_core.cu
@@ -19,7 +19,7 @@ extern "C" void compat_usleep(uint64_t waitTime)
             HANDLE timer;
             LARGE_INTEGER ft;
 
-            ft.QuadPart = -10*(LONGLONG)waitTime; // Convert to 100 nanosecond interval, negative value indicates relative time
+            ft.QuadPart = -10ll * int64_t(waitTime); // Convert to 100 nanosecond interval, negative value indicates relative time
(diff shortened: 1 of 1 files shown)


awhittle3:
If you'd like to support, consider donating xmr to 44htu1Ti385jiqWZx8duxcBmP9zvwcuHVKmwnCQSzcHoX7A4jdvwH7R5ofkg4ESgMDJTHfqwcSNwm55jSwfnDokqHQCQjdc

fireice-uk:
Oh crap, nice catch. There is a simpler way to write it though:

>  ft.QuadPart = -10ll * int64_t(waitTime);

This is better because it doesn't have new programmers scratching their heads about cast orders. How many times did you see  `((unsigned)i) + 5` in C :)?

fireice-uk:
We don't give out bounties for one liners unfortunately =)

awhittle3:
@fireice-uk No worries, just happy to contribute :)

fireice-uk:
Actually - I think I screwed up =) -> I told you to use C++ semantics in an extern "C" - I will let @psychocrypt direct this one.

fireice-uk:
If depending on what he thinks will work, he might force push to your branch. You don't need to do anything :)

psychocrypt:
thx for fixing. C++ is fine, we can remove extern C here but should do this in a separate PR.

PR #452, opened 2017-12-09
Something minor

taisel:
Probably needs to be vetted because tired me did this, but I was trying to explore the code and tried merging some buffers.


taisel:
Also did this when y'all updated the code, so there might be something amiss now. Again, tired me explorin' the code.

taisel:
https://github.com/fireice-uk/xmr-stak/commit/efc8dc447adc8fcbf2c1faf0de2dacf8d38889ac#diff-5b1d0c2b2acc529d2838a0b2eb26b3c9

Yeah, we've got two different recent passes on the same file fightin'.

taisel:
I just prefer using a multidimensional array over fancy pointer arith, as the hardcoded array offsets can be often baked into the machine code directly, rather than reside in data.

taisel:
I can see what y'all did to optimize as well. Re-ordering the XOR sequencies to make the AES local buffer reads more sequential.

psychocrypt:
Thanks for you PR. Pleas reopen the PR against the `dev` branch. https://github.com/fireice-uk/xmr-stak/blob/master/CONTRIBUTING.md

Is this change giving use any hasr rate increase or is this only a style change?

PR #454, opened 2017-12-09
Coalesce AES local buffers + Hueristics change

taisel:
Moving PR around...


taisel:
Haven't been able to pin down the numbers yet for any hashrate changes. Any delta seems to be within the usual big cloudy margin of error.

Theoretically though the codegen *should* be cleaner.

taisel:
Still checking to make sure I didn't mess something up.

taisel:
ok there we go.

taisel:
Built it myself to confirm it does run, it's reporting at least reporting the same numbers as before, perf delta might be well below the noise threshold.

psychocoderHPC:
could you please performance number with and without this PR. Please do not activate cpu mining during the tests and use only one GPU.

psychocrypt:
Could you please post performance number of PR and without this PR. Please deactivate CPU mining and use only one GPU (1min avg is ok).

taisel:
https://imgur.com/a/xsXi3

taisel:
Oof, that's an extra change. It's basically trying to do what people have been empirically discovering on r/moneromining. Irregardless, it seems to have taken pog champ on the rx 460 hash rate when determining the best config.

taisel:
The pull454 result in the image album above is also compiled against the latest change for heuristics. Seems to work for the rx 460 at least. 454 h/s isn’t bad on it.

psychocrypt:
Please do not mix to orthogonal changes within one pull request. Please put the auto suggestion change into an separate PR. 

Never the I will add some comments you can edit for the extra auto suggestion PR.

xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -97,34 +95,46 @@ class autoAdjust
...
-			// keep 128MiB memory free (value is randomly chosen)
-			size_t availableMem = ctx.freeMem - (128u * byteToMiB);
-			// 224byte extra memory is used per thread for meta data
+			/*
+			  GCN has a hard architectural limit of 64 CUs max (Invalid for other archs though).
+			  Reserve 1 CU for the system. 
+			*/
+			size_t maxComputeUnitsAvailable = std::max(std::min(ctx.computeUnits, 64) - 1, 1);
+			//Each CU is 64 threads executed in an SIMT fashion:
+			size_t maxThreadsAvailable = maxComputeUnitsAvailable << 6;
```
psychocrypt: Please not use magic bit shift, try to keep the code human readable. `X << 6` is equal to `X * 64` the compiler will replace it with a bit shift.

xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -97,34 +95,46 @@ class autoAdjust
...
-
-			// keep 128MiB memory free (value is randomly chosen)
-			size_t availableMem = ctx.freeMem - (128u * byteToMiB);
-			// 224byte extra memory is used per thread for meta data
+			/*
+			  GCN has a hard architectural limit of 64 CUs max (Invalid for other archs though).
+			  Reserve 1 CU for the system. 
+			*/
+			size_t maxComputeUnitsAvailable = std::max(std::min(ctx.computeUnits, 64) - 1, 1);
+			//Each CU is 64 threads executed in an SIMT fashion:
```
psychocrypt: A wavefront on a AMD gpu is 64 threads. This not means that you should only use 64 threads per multiprozessor. If you oversubscribe the multiprocessor you can hide memory latencies.

xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -97,34 +95,46 @@ class autoAdjust
...
+			size_t maxThreadsAvailable = maxComputeUnitsAvailable << 6;
+			//Keep 128MiB memory free (value is randomly chosen):
+			size_t availableMem = ctx.freeMem - (128u << 20);
```
psychocrypt: please keep the human readable variable `byteToMiB`

xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -97,34 +95,46 @@ class autoAdjust
...
+			//224byte extra memory is used per thread for meta data
 			size_t perThread = hashMemSize + 224u;
 			size_t maxIntensity = availableMem / perThread;
-			size_t possibleIntensity = std::min( maxThreads , maxIntensity );
-			// map intensity to a multiple of the compute unit count, 8 is the number of threads per work group
-			size_t intensity = (possibleIntensity / (8 * ctx.computeUnits)) * ctx.computeUnits * 8;
-			conf += std::string("  // gpu: ") + ctx.name + " memory:" + std::to_string(availableMem / byteToMiB) + "\n";
+			size_t possibleIntensity = std::min(maxThreadsAvailable, maxIntensity);
+			//Alias intensity against the smallest work group size possible:
+			size_t intensity = (possibleIntensity >> 3) << 3;
```
psychocrypt: some here please keep it readable

xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -97,34 +95,46 @@ class autoAdjust
...
+			conf += std::string("  // gpu: ") + ctx.name + " memory:" + std::to_string(availableMem >> 20) + "\n";
 			conf += std::string("  // compute units: ") + std::to_string(ctx.computeUnits) + "\n";
-			// set 8 threads per block (this is a good value for the most gpus)
-			conf += std::string("  { \"index\" : ") + std::to_string(ctx.deviceIdx) + ",\n" +
+			//set 8 threads per block (this is a good value for the most gpus)
+			//Create two instances if more than one CU is available:
+			if (maxComputeUnitsAvailable > 1)
+			{
+				//Reports seem to indicate splitting a physical device into two works better:
+				conf += std::string("  { \"index\" : ") + std::to_string(ctx.deviceIdx) + ",\n" +
```
psychocrypt: please move the config generation after the if condition and collect all needed states into helper variables.

xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -97,34 +95,46 @@ class autoAdjust
...
+				"    \"intensity\" : " + std::to_string((intensity >> 7) << 6) + ", \"worksize\" : " + std::to_string(8) + ",\n" +
+				"    \"affine_to_cpu\" : false, \"strided_index\" : false\n" +
+				"  },\n";
+				intensity -= (intensity >> 7) << 6;
```
psychocrypt: keep is readable

psychocrypt:
If you use in your local git fork a branch per feature you can work in independent features at the same time.
This will allow to follow the [successful git branching model](http://nvie.com/posts/a-successful-git-branching-model/)

xmrstak/backend/amd/autoAdjust.hpp
taisel: Totally agree. 

I just found empirically that cryptonight is the odd man out here in that any form of context swapping seems to add load to the memory bottleneck. I should probably have been clearer in the comment I’m explicitly trying to avoid the context switching, even if nearly free (10 wavefronts can be buffered per SIMD unit in each CU I think...).

fireice-uk:
> I also wasn’t sure how you’d respond to using bitshifts, but I guess keeping it expanded or simple is preferred.

`constexpr size_t d = 1024;`
`x /= d`

and

`x >>= 10`

compile to exactly the same code.

candrews:
I believe this PR (and any others) should be against the `dev` branch too (please correct me if I'm misunderstanding the xmr-stak development process).

fireice-uk:
@candrews You are misreading - master is the source branch

xmrstak/backend/amd/amd_gpu/gpu.cpp
```diff
@@ -522,7 +522,13 @@ std::vector<GpuContext> getAMDDevices(int index)
...
 			// if environment variable GPU_SINGLE_ALLOC_PERCENT is not set we can not allocate the full memory
 			ctx.deviceIdx = k;
-			ctx.freeMem = std::min(ctx.freeMem, maxMem);
+			/*
+			* System does __not always__ report true memory left, 
+			* only the max amount we can allocate at once...
+			* It is a __hack__ to not be querying the max alloc before each
+			* and every alloc if trying to figure out the amount available.
+			*/
+			ctx.freeMem = maxMem * std::floor(ctx.freeMem / maxMem);
```
taisel: I found out why Vega users were reporting less than 4GB being reported in amd.txt.

https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clGetDeviceInfo.html
CL_DEVICE_MAX_MEM_ALLOC_SIZE
Max size of memory object allocation in bytes. The minimum value is max (1/4th of CL_DEVICE_GLOBAL_MEM_SIZE, 128*1024*1024)

The gotcha is that the allocation size reported does not have to be the max mem size. This commit here is a HACK and needs to be replaced with something different. The proper code might need to query the devices at runtime than ahead of time!

xmrstak/backend/amd/autoAdjust.hpp
taisel: The ">> 7"'s have a purpose - they round down to units of 64 threads for the dual thread optimization. The << 6's though have been changed back to * 64. I reversed the multiply and divide by one megabyte change.

xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -97,34 +97,46 @@ class autoAdjust
...
+			//set 8 threads per block (this is a good value for the most gpus)
+			//Create two instances if more than one CU is available:
+			if (maxComputeUnitsAvailable > 1)
+			{
+				//Reports seem to indicate splitting a physical device into two works better:
+				conf += std::string("  { \"index\" : ") + std::to_string(ctx.deviceIdx) + ",\n" +
+				"    \"intensity\" : " + std::to_string((intensity >> 6) * 32) + ", \"worksize\" : " + std::to_string(8) + ",\n" +
+				"    \"affine_to_cpu\" : false, \"strided_index\" : false\n" +
+				"  },\n";
+				intensity -= (intensity >> 6) * 32;
```
taisel: We're cutting off the bottom 6 bits to round to the nearest 64th. Then we multiply by 32 so we get half of the total shaders available, rounded down, for the first gpu thread specified.

taisel: That's done as mentioned elsewhere the system likes to reserve a whole CU.

taisel:
Last commit shows why vega users had a barrier around 2k intensity per thread. Namely the system reports only just under 4GB as the max you can allocate, even on fresh boot. 

The issue is that max allocation possible that is reported does not have to be the free space remaining. Lame

taisel:
Also https://github.com/fireice-uk/xmr-stak/commit/6274190f3a8b8b03d42d7065447d5ec7b16e3d39

Needs eyeballs.

psychocrypt:
@taisel please separate the independent changes. With the new possible opencl memory fix you opened the third changes into this pull request. It is not possible to discuss the changes if we mix it.

psychocrypt:
@taisel The last commit will not solve that we can't set 2k for the intensity. The point is that you changed the auto suggestion. The auto suggestion is only a try to get a stable good config. To get the last amount of hash card out of the card you can still change the config by hand in the file `amd.txt`. The intensity changed by hand is independent of the auto suggestion. But less use discuss that in an separate pull request.

taisel:
The commits were to avoid it auto suggesting too high a value while working around Vegas having a top low auto suggest (because the new logic splits it up into two thread).

taisel:
I'll close this for now. I won't be able to divide it up soonish, so I'll be maintaining my fork until I get the time to split up the individual pieces one by one and code style review.

PR #464, opened 2017-12-10
handle cuda error codes

psychocrypt:
handle all error codes from the cuda API calls.


fireice-uk:
Somehow I forgot to merge this one, sorry.

PR #613, opened 2017-12-18
add OpenCL compatibility mode

psychocrypt:
- add new option `comp_mode` to the amd config
- disable if guards within opencl kernel if `comp_mode : false`

# Why this new option?
Many users report that the hash rate with the amd miner before the unified version was better than now. This PR will allow to remove the in kernel guards which allows to call a kernel with a intensity which is not a multiple of the worksize. Maybe the ifs in the kernel prevent the compile to optimize a few lines of code. This PR should show if there is a improve available without the if guards.

# HowTo
- compile the miner
- remove/rename your old amd.txt
- start the miner (a new amd.txt is created)
- check the hash rate (must be equal to the current release)
- change `comp_mode` in amd.txt to false (hope that the `intensity` is a multiple of the `worksize`)
- run the miner and compare the hash rate to the previous run

# Important 
~~- [x] this PR is a WIP/ please do not merge~~

# Tests
- [x] tests with AMD RX570

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/gpu.cpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.cpp
@@ -332,8 +332,8 @@ size_t InitOpenCLGpu(cl_context opencl_ctx, GpuContext* ctx, const char* source_
 
 	char options[256];
 	snprintf(options, sizeof(options), 
-		"-DITERATIONS=%d -DMASK=%d -DWORKSIZE=%llu -DSTRIDED_INDEX=%d -DMEM_CHUNK=%d",
-		hasIterations, threadMemMask, int_port(ctx->workSize), ctx->stridedIndex, int(1u<<ctx->memChunk));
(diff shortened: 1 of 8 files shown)


psychocrypt:
@JerichoJones @davidpesce  could you please check this pull request if you get more hashes with `comp_mode : false`

JerichoJones:
Vega56
Intensity: 1932
Blocks: 8
It doesn't seem to make a real difference. It looks like I went from 986H/s per thread (TRUE) to 988H/s per thread (FALSE). ~0.2% increase? (I'm terrible at math)
I did try increasing/reducing Intensity by 4 but it just got worse.

davidpesce:
@JerichoJones do you have both your threads at 1932? I'm running both threads at 1850. 

@psychocrypt didn't get any noticeable increase

psychocrypt:
thx for testeing, it looks like that this PR will not increase the hash rate, therefore I will keep it in the WIP state and first test some other things.

psychocrypt:
@fireice-uk I updated this PR to be merge able. This PR will only give a few hashes more but better than nothing^^

psychocrypt:
I rebased this branch to solve the conflicts

PR #625, opened 2017-12-20
Fix CUDA 9.1 compatibility issue

vicnicius:
This fixes the `fatal error: device_functions.hpp: "No such file or directory" <device_functions.hpp>` issue when compiling on Ubuntu. 

It seems related to the new CUDA version.

Let me know if there's need to first open an issue or something like this.


vicnicius:
Closing it as some fixes were already merged to `dev` branch. Will try to compile against it. :+1:

PR #649, opened 2017-12-21
Stop UAC popup for elevated ctx

fireice-uk:
Very quick fix - haven't even built it yet.


psychocrypt:
looks good on WINDOWS 10.
I will run also a test on windows 7 tomorrow.

PR #670, opened 2017-12-22
Rearrange prefetch for slight performance increase in CPU miner

dataway:
I moved the two prefetch commands to take place as early as possible to give the cache time to fetch from slow memory. On my mediocre CPU (i7-3770S) this increases the hash rate by about 1%. The effect is probably more noticable for CPUs with small cache.


fireice-uk:
@dataway Can you configure your email properly in git and squash this commit?

dataway:
What do you mean exactly?

fireice-uk:
You commits are not recognised as yours by github, You will need to close this pr, set your email to `uk@dataway.ch` from `Anthony Uk uk@dataway.ch` then finally checkout the dev branch and re-commit the changes.

When you do that make sure there are no merge-commits in the pr (you have two there)

PR #697, opened 2017-12-25
Updated config.tpl comments to include how to exclude CPU/GPUs

dangrabbits:
This is my first time attempting to use github. I think I'm trying to commit to the dev branch but I am not knowledgeable enough to be certain. Sorry if it's wrong.


psychocrypt:
Do you tested if it is possible to set all arrays to `null`?

psychocrypt:
ping

PR #698, opened 2017-12-25
Update donate-level.hpp

issyaboi:


psychocrypt:
I think thats a bad idea

PR #705, opened 2017-12-25
Rearrange prefetch for slight performance increase in CPU miner

dataway:
I moved the two prefetch commands to take place as early as possible to give the cache time to fetch from slow memory. On my mediocre CPU (i7-3770S) this increases the hash rate by about 1%. The effect is probably more noticable for CPUs with small cache.

Before this patch 277.7 H/s, afterwards 280.8 H/s

This is a resubmission of #670 which I messed up.

### DIFF
--- a/xmrstak/backend/cpu/crypto/cryptonight_aesni.h
+++ b/xmrstak/backend/cpu/crypto/cryptonight_aesni.h
@@ -317,10 +317,9 @@ void cryptonight_hash(const void* input, size_t len, void* output, cryptonight_c
 
 		_mm_store_si128((__m128i *)&l0[idx0 & MASK], _mm_xor_si128(bx0, cx));
 		idx0 = _mm_cvtsi128_si64(cx);
-		bx0 = cx;
-
 		if(PREFETCH)
 			_mm_prefetch((const char*)&l0[idx0 & MASK], _MM_HINT_T0);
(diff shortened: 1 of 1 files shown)


psychocrypt:
Do you set `no_prefetch` to false for your tests?

dataway:
no_prefetch is true but it still improved.

dataway:
Tested with no_prefetch to false. Also see slight improvement

psychocrypt:
I will check this PR in a week, currently I am busy

JerichoJones:
CPU Only
UAC ON
Memory pinned

Dual Xeon x5650 - "no_prefetch" : true
PR #705 484 H/s 60 sec
v2.1 486 H/s 60 sec

Dual Xeon x5650 - "no_prefetch" :false
PR #705 461 H/s 60 sec
v2.1 460 H/s 60 sec

Ryzen 1700x OC 3.9 Ghz - "no_prefetch" : true
PR #705 646 H/s 60 sec
v2.1 646 H/s 60 sec

Ryzen 1700x OC 3.9 Ghz - "no_prefetch" : false
PR #705 637 H/s 60 sec
v2.1 631 H/s 60 sec

fireice-uk:
Looks like gains are purely psychological - but will approve it - some compilers may not be clever enough to do that themselves.

PR #740, opened 2017-12-29
remove branch differgence in OpenCL code

psychocrypt:
Remove branch divergences in AMD OpenCl code based on #454.
The effect of this PR should be a very small hash rate increase.

difference to #454
- Instead of using the OpenCl select statement I used short if which should be translated to the equal code. 
- this PR is not using unsave casts to `uint` for pointer

@JoKeRz42o @JerichoJones could you please test this PR?

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
+++ b/xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
@@ -653,21 +653,11 @@ __kernel void cn2(__global uint4 *Scratchpad, __global ulong *states, __global u
 
 			for(int i = 0; i < 25; ++i) states[i] = State[i];
 
-			switch(State[0] & 3)
-			{
-				case 0:
-					Branch0[atomic_inc(Branch0 + Threads)] = get_global_id(0) - get_global_offset(0);
-					break;
-				case 1:
-					Branch1[atomic_inc(Branch1 + Threads)] = get_global_id(0) - get_global_offset(0);
-					break;
-				case 2:
-					Branch2[atomic_inc(Branch2 + Threads)] = get_global_id(0) - get_global_offset(0);
-					break;
-				case 3:
-					Branch3[atomic_inc(Branch3 + Threads)] = get_global_id(0) - get_global_offset(0);
-					break;
-			}
+			ulong StateSwitch = State[0] & 3;
+			__global uint *destinationBranch1 = StateSwitch == 0 ? Branch0 : Branch1;
+			__global uint *destinationBranch2 = StateSwitch == 2 ? Branch2 : Branch3;
+			__global uint *destinationBranch = StateSwitch < 2 ? destinationBranch1 : destinationBranch2;
+			destinationBranch[atomic_inc(destinationBranch + Threads)] = gIdx;
 		}
 	}
 	mem_fence(CLK_GLOBAL_MEM_FENCE);
@@ -704,8 +694,7 @@ __kernel void Skein(__global ulong *states, __global uint *BranchBuf, __global u
 
 		for(uint i = 0; i < 4; ++i)
 		{
-			if(i < 3) t[0] += 0x40UL;
-			else t[0] += 0x08UL;
+			t[0] += i < 3 ? 0x40UL : 0x08UL;
 
 			t[2] = t[0] ^ t[1];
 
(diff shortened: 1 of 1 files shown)


xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
```diff
@@ -761,22 +749,37 @@ __kernel void JH(__global ulong *states, __global uint *BranchBuf, __global uint
...
+			 * else
+			 * {
+			 *     input[7] = 0x4006000000000000UL;
+			 *     for(int x = 0; x < 7; ++x) input[x] = 0x00UL;
+			 * }
+			 */
+
+			const bool isLessThree = i < 3;
+			const int shifted = i << 3;
+			input[0] = isLessThree ? (states[shifted]) : 0x00UL;
```
fireice-uk: Unless benchmarked  or otherwise confirmed to be optimised by the compiler, ?: operator is just another way to write if / else.

Instead:

Unroll the `for(int i = 0; i < 5; ++i)` loop - it is a stupid aggregate of different tasks anyway. The whole thing can be rewritten as:

2x

```
#pragma unroll 8
for(int x = 0; x < 8; ++x) input[x] = (states[(i << 3) + x]);
<epilogue - possibly inlined function>
```

1x

```
input[0] = (states[24]);
input[1] = 0x80UL;
#pragma unroll 6
for(int x = 2; x < 8; ++x) input[x] = 0x00UL;
<epilogue - possibly inlined function>
```

3x

````
input[7] = 0x4006000000000000UL;
#pragma unroll 7
for(int x = 0; x < 7; ++x) input[x] = 0x00UL;
<epilogue - possibly inlined function>
````

psychocrypt: No if is not equal to `cond ? :`. The short version is a select option. The if can be transformed into the short versions but it strongly depends on the compiler. The short version is translated into hardware select operations which are available on the gpus.

I do not understand what you mean with the three different code blocks. How should they avoid the branch divergence?

psychocrypt: Please keep in mind that unrolling loops with to high numbers (e.g. 8) will take a high number of registers and therefore can slow downthe kernel.

JerichoJones:
Do you still want this tested? If so, can you post a source zip. At this point, Git makes me feel bad about myself. :(

psychocrypt:
yes please test this pr
https://github.com/psychocrypt/xmr-stak/archive/topic-tuneAmd.zip

xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
fireice-uk: @psychocrypt the code that i posted above can literally replace the `for(int i = 0; i < 5; ++i)`

psychocrypt: Yes I can join the hand unrilled code to a loop. I will do it later.

JerichoJones:
No issues with AMD & CPU... do you need NVIDIA as well.

psychocrypt:
no this change only amd code. Do you see a hash rate increase?

JerichoJones:
Not on my RX550s.
I'll try my Vega64.

JerichoJones:
1-2 hashes faster per thread.

psychocrypt:
thx thats in the expected range

psychocrypt:
@fireice-uk I rewrote the code part, it was not possible to use a inline function because of the used macros inside the algorithm.

psychocrypt:
@fireice-uk This PR is ready to merge

taisel:
The select builtin is full of gotchas. The cast to uint wasn't originally, intended, it's just that the select builtin capabilities exposed by the AMD OpenCL compiler is very... limited.

For instance, I was code golfing in another part to convert a branch to select... I present to you the retardation that is https://github.com/taisel/xmr-stak/commit/2fdf5b8ce8708cc8dee626392c11c185d4c13be0

I use the select statement as it has a corresponding instruction in the ISA. Some data types aren't supported by the select statement... like what was attempt earlier... and I don't like leaving that up to the compiler to screw up when pattern matching an if block to a conditional move.

PR #754, opened 2017-12-30
Modify invalid result report to show GPU id

dougvj:
I was unable to tell which GPU was creating invalid hashes. This modification of the error report fixes that.


fireice-uk:
Thanks! Will review it after the new-year hangover passes ^^

SplitfingerJoe:
Confirmed to be working. Very useful, thank you!
One more suggestion: it would be great to have a logfile for errors, or maybe to include hardware error messages in verbose level 1. This would make it much easier to identify faulty hardware/too high overclocks on specific cards.

Issue #490, opened 2017-12-11
HD 6970 Error Buffer Size or reduced HashRate


jahknem:
When running the miner with my HD 6970 it says: "Error CL_INVALID_BUFFER_SIZE when calling clCreateBuffer to create hash scratchpads buffer."
This happens if I use either 0 or anything above 256 as intensity variable.
When using anything between 0 and 256 as Intensity the Hashrate is still below what it should be at max 100 H/s.
After that it only starts the CPU Miner (albeit at a very reduced hashrate. 80H/s instead of the 200 H/s it achieves with the xmrig 2.4.3)

# Basic information
  - Type of the CPU: FX 6300
  - Type of the GPU: HD 6970

# Compile issues
  - Which OS do you use?
  Windows Server 2012 R2 64 Bit
  add **all** commands you used and the **full** compile output here
  ```
  ```
  run `cmake -LA .` in the build folder and add the output here
  ```
  ```
# Issue with the execution
  - Do you compiled the miner by our own?
  No. I used the precompiled binary v2.1.0

  run `./xmr-stak --version-long` and add the output here
  ```
"Version: xmr-stak/2.1.0/30888a3/master/win/nvidia-amd-cpu/aeon-monero/20
Press any key to exit."
  ```
# AMD OpenCl issue

  run `clinfo` and add the output here
  ```
Microsoft Windows [Version 6.3.9600]
(c) 2013 Microsoft Corporation. Alle Rechte vorbehalten.

C:\Users\Jan>clinfo
Number of platforms:                             1
  Platform Profile:                              FULL_PROFILE
  Platform Version:                              OpenCL 2.0 AMD-APP (1800.8)
  Platform Name:                                 AMD Accelerated Parallel Proces
sing
  Platform Vendor:                               Advanced Micro Devices, Inc.
  Platform Extensions:                           cl_khr_icd cl_khr_d3d10_sharing
 cl_khr_d3d11_sharing cl_khr_dx9_media_sharing cl_amd_event_callback cl_amd_offl
ine_devices


  Platform Name:                                 AMD Accelerated Parallel Proces
sing
Number of devices:                               2
  Device Type:                                   CL_DEVICE_TYPE_GPU
  Vendor ID:                                     1002h
  Board name:                                    AMD Radeon HD 6900 Series
  Device Topology:                               PCI[ B#1, D#0, F#0 ]
  Max compute units:                             24
  Max work items dimensions:                     3
    Max work items[0]:                           256
[...]

Issue #81, opened 2017-10-30
Auto configuration first start- AMD GPU "intensity" : 0


deereslv:
Running latest xmr-stak-dev, on linux with rx480( not OC), compile without problems, cpu backend i5  works flawlessly , but having problem with GPU.
After first start, and auto configuration, in amd.txt, suggested configuration is: 
"intensity" : 0, "worksize": 8
And if i increase intensity more than 32, get many "AMD invalid result" errors. 
Also tried playing with worksize , but no difference. 
Sometimes most errors are upon miner start, and sometimes start without errors, and after some time starts throwing errors

RESULT REPORT
Difficulty       : 100001
Good results     : 316 / 403 (73.8 %)
Avg result time  : 100.1 sec
Pool-side hashes : 2958631

Top 10 best results found:
|  0 |         17327791 |  1 |          6890190 |
|  2 |          5178865 |  3 |          4649874 |
|  4 |          4548370 |  5 |          3991239 |
|  6 |          3690780 |  7 |          3579746 |
|  8 |          3028025 |  9 |          2505256 |

Error details:
| Count | Error text                       | Last seen           |
|    87 | AMD Invalid Result               | 2017-10-30 07:49:40 |

Note, tried different GPU's, all rx480 and almost new, i took them from my eth rig (works without problems) to test aeon mining. Same configuration both my rig and testing computer,  ubuntu/drivers, so GPU should be fine

deereslv:
CLINFO:

$ clinfo 

Number of platforms:				 1
  Platform Profile:				 FULL_PROFILE
  Platform Version:				 OpenCL 2.0 AMD-APP (1800.11)
  Platform Name:				 AMD Accelerated Parallel Processing
  Platform Vendor:				 Advanced Micro Devices, Inc.
  Platform Extensions:				 cl_khr_icd cl_amd_event_callback cl_amd_offline_devices 


  Platform Name:				 AMD Accelerated Parallel Processing
Number of devices:				 1
  Device Type:					 CL_DEVICE_TYPE_GPU
  Vendor ID:					 1002h
  Board name:					 AMD Radeon (TM) RX 480 Graphics
  Device Topology:				 PCI[ B#2, D#0, F#0 ]
  Max compute units:				 36
  Max work items dimensions:			 3
    Max work items[0]:				 256
    Max work items[1]:				 256
    Max work items[2]:				 256
  Max work group size:				 256
  Preferred vector width char:			 4
[...]

PR #758, opened 2017-12-31
Add warning and fallback when auto intensity is 0

dougvj:
Occassionally the auto adjust doesn't find enough memory and the
intensity is detected too low and aligned to 0 with the compute units.
This patch fixes this situation by issuing a warning with a suggestion
to set environment vars and then ignoring the alignment to 0

Per several issues:

Principally:
https://github.com/fireice-uk/xmr-stak/issues/81

Related:
https://github.com/fireice-uk/xmr-stak/issues/490
https://github.com/fireice-uk/xmr-stak/issues/472

### DIFF
--- a/xmrstak/backend/amd/autoAdjust.hpp
+++ b/xmrstak/backend/amd/autoAdjust.hpp
@@ -94,7 +94,6 @@ private:
 		}
 
 		std::string conf;
-		int i = 0;
 		for(auto& ctx : devVec)
 		{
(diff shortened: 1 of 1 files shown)


psychocrypt:
I will check this soon, maybe I will push some change on top of you PR to remove the link to all issue. I will also check first if it is better that no config is generated in the case where the intensity is zero.

psychocrypt:
@fireice-uk  Could you please check this PR, I pushed a commit on top and therefore will give the review into your hands.

PR #772, opened 2018-01-02
It compiles on OpenBSD/adJ 6.2

### LINKED ISSUE #771: Compilation on OpenBSD/adJ 6.2
# Basic information
  -  Type of CPU: amd64

# Compile issues
  - Operating system: OpenBSD/adJ 6.2

It is possible to start building with:
```sh
cmake . -DCUDA_ENABLE=OFF -DOpenCL_ENABLE=OFF -DMICROHTTPD_ENABLE=OFF -DHWLOC_ENABLE=OFF
```
But runing `make` produces:
```
[ 44%] Building CXX object CMakeFiles/xmr-stak-backend.dir/xmrstak/backend/cpu/minethd.cpp.o
/home/vtamara/comp/monero/xmr-stak/xmrstak/backend/cpu/minethd.cpp:88:2: error: unknown type name 'cpu_set_t'
        cpu_set_t mn;
        ^
/home/vtamara/comp/monero/xmr-stak/xmrstak/backend/cpu/minethd.cpp:91:42: error: use of undeclared identifier 'cpu_set_t'
        return pthread_setaffinity_np(h, sizeof(cpu_set_t), &mn) == 0;
                                                ^
2 errors generated.
```

vtamara:
Fixes #771 

After compiling it can run, but only one thread (I will open an issue about it later).

### DIFF
--- a/xmrstak/backend/cpu/minethd.cpp
+++ b/xmrstak/backend/cpu/minethd.cpp
@@ -84,7 +84,7 @@ bool minethd::thd_setaffinity(std::thread::native_handle_type h, uint64_t cpu_id
 	CPU_ZERO(&mn);
 	CPU_SET(cpu_id, &mn);
 	return pthread_setaffinity_np(h, sizeof(cpuset_t), &mn) == 0;
-#else
+#elif !defined(__OpenBSD__)
 	cpu_set_t mn;
 	CPU_ZERO(&mn);
 	CPU_SET(cpu_id, &mn);
--- a/xmrstak/backend/cpu/minethd.cpp
+++ b/xmrstak/backend/cpu/minethd.cpp
@@ -84,7 +84,9 @@ bool minethd::thd_setaffinity(std::thread::native_handle_type h, uint64_t cpu_id
 	CPU_ZERO(&mn);
 	CPU_SET(cpu_id, &mn);
 	return pthread_setaffinity_np(h, sizeof(cpuset_t), &mn) == 0;
-#elif !defined(__OpenBSD__)
+#elif defined(__OpenBSD__)
+        printer::inst()->print_msg(L0,"WARNING: thread pinning is not supported under OPENBSD.");
+#else
 	cpu_set_t mn;
 	CPU_ZERO(&mn);
 	CPU_SET(cpu_id, &mn);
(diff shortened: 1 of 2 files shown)


xmrstak/backend/cpu/minethd.cpp
```diff
@@ -84,7 +84,7 @@ bool minethd::thd_setaffinity(std::thread::native_handle_type h, uint64_t cpu_id
 	CPU_ZERO(&mn);
 	CPU_SET(cpu_id, &mn);
 	return pthread_setaffinity_np(h, sizeof(cpuset_t), &mn) == 0;
-#else
+#elif !defined(__OpenBSD__)
```
psychocrypt: Could you please try to change line 62 to 
`#elif defined(__FreeBSD__) || defined(__OpenBSD__)`
and than remove this line. This should solve #771 without removing the thread pinning feature.

psychocrypt (changes requested):
Thanks for the PR, could you please try the annotated changes. The error in #771 looks like an missing include which can be solved by adding the define in line 62.

xmrstak/backend/cpu/minethd.cpp
vtamara: I did, however it produces:
```
/home/vtamara/comp/monero/xmr-stak/xmrstak/backend/cpu/minethd.cpp:88:2: error:
      unknown type name 'cpu_set_t'
        cpu_set_t mn;
        ^
/home/vtamara/comp/monero/xmr-stak/xmrstak/backend/cpu/minethd.cpp:91:42: error:
      use of undeclared identifier 'cpu_set_t'
        return pthread_setaffinity_np(h, sizeof(cpu_set_t)...
                                                ^
2 errors generated.
```

What I have found about CPU affinity in OpenBSD up to now are these commits from 2009:
<https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/sys/proc.h.diff?r1=1.112&r2=1.113&f=h>
<https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/sys/proc.h.diff?r1=1.115&r2=1.116&f=h>
they introduce the structure `struct cpuset` and some operations on them, however up to OpenBSD 6.2 there is no function `pthread_setaffinity_np`.

psychocrypt: OK in this case could you please add a extra code path for OPENBSD and show the user a WARNING that thread pinning is not supported under OPENBSD.
```
printer::inst()->print_msg(L0,"WARNING: thread pinning is not supported under OPENBSD.");
```

vtamara: Done

psychocrypt (approved):

psychocrypt:
thx for your work

PR #778, opened 2018-01-02
Rename "MacOS" to "macOS"

b-:
Apple has a silly way of capitalizing things, but it's still part of the name.

Basically this pull request is just `s/MacOS/macOS/g`

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ XMR-Stak is a universal Stratum pool miner. This miner supports CPUs, AMD and NV
 ## Features
 
 - support all common backends (CPU/x86, AMD-GPU and NVIDIA-GPU)
-- support all common OS (Linux, Windows and MacOS)
+- support all common OS (Linux, Windows and macOS)
 - supports algorithm cryptonight for Monero (XMR) and cryptonight-light (AEON)
 - easy to use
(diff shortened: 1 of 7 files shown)


Raginglikeaboss:
Good catch.  I agree it may seem silly if Apple but I believe this pull request show be quickly incorporated.  It helps the final product be more refined and polished.

fireice-uk:
@b- Can you squash this pr to a single commit?

b-:
I know this is a total newbie question I shouldn’t have to ask here, but how do I squash commits? I won’t have to undo everything and then redo it, will I? :/

XhmikosR:
@fireice-uk: you can choose to squash when merging, https://github.com/blog/2141-squash-your-commits

fireice-uk:
@XhmikosR 

Last time I tried to skirt around having merge commits I broke the whole repo :D. I wouldn't do it on a dev or stable. 

@b- 

Those commands (run them in the project dir) should do the trick

```
git reset --soft HEAD~6
git commit --amend
git push -f
```

The last one is destructive, so preferably verify that everything is ok before running it. You can use ```git log``` to view the new commit history.

b-:
boom

fireice-uk:
Note to self: CI fails here can be ignored.

Issue #125, opened 2017-11-16
Hash Rate Display Enhancement


JerichoJones:
Add hashrate totals per hardware type so we don't have to take out the calculator to see how a particular component is doing (CPU is really culprit with a large number of cores).

psychocrypt:
I will add it after the release. Today we will do a feature freez and
branch out the release candidate.

ChaosMarc:
@psychocrypt this issue is fixed by now and can be closed

PR #803, opened 2018-01-05
#125: separate hashrate sums for each component

ChaosMarc:
As in #125 suggested each mining component (CPU, GPU) now displays an individual sum in the hashrate report. I also changed this display of sums to show "0.0" instead of "(na)" if one of the summed up components shows (na)

report


psychocrypt:
Thanks for this nice PR, I will check it soon.

fireice-uk:
@psychocrypt keep in mind that I'm working on total conversion of that code to a template system. I will let you make the call as to whether to merge this one in the meantime.

psychocrypt:
@fireice-uk I assigned it to you that you can handle it. This will avoid issues with you current work.

PR #824, opened 2018-01-07
- Exposed port 1600, in order view html reports.

doublebyte1:


fireice-uk:
Hi. The port is user-configured, so 1600 is not necessarily the port that other users will use. Please feel free to reopen with something that is suitable for everyone.

PR #832, opened 2018-01-08
Optimize Skein

taisel:
- Eliminate modulus math (It runs slow inside microcode).
- Convert whatever the hell was going on into a rotate op. Like... Someone kinda reinvented the wheel in order to do a simple rotate.

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/opencl/wolf-skein.cl
+++ b/xmrstak/backend/amd/amd_gpu/opencl/wolf-skein.cl
@@ -3,6 +3,7 @@ R"===(
 #define WOLF_SKEIN_CL
 
 // Vectorized Skein implementation macros and functions by Wolf
+// Updated by taisel
 
 #define SKEIN_KS_PARITY	0x1BD11BDAA9FC1A22
 
--- a/xmrstak/backend/amd/amd_gpu/opencl/wolf-skein.cl
+++ b/xmrstak/backend/amd/amd_gpu/opencl/wolf-skein.cl
@@ -30,10 +30,19 @@ static const __constant ulong SKEIN512_256_IV[8] =
 	p.s7 += q; \
 } while(0)
 
+ulong SKEIN_ROT(const uint2 x, const uint y)
+{
+	if(y < 32) return(as_ulong(amd_bitalign(x, x.s10, 32 - y)));
+	else return(as_ulong(amd_bitalign(x.s10, x, 32 - (y - 32))));
+}
+
 void SkeinMix8(ulong4 *pv0, ulong4 *pv1, const ulong4 rc)
 {
 	*pv0 += *pv1;
-	*pv1 = rotate(*pv1, (ulong4)rc);
+	(*pv1).s0 = SKEIN_ROT(as_uint2((*pv1).s0), rc.s0);
+	(*pv1).s1 = SKEIN_ROT(as_uint2((*pv1).s1), rc.s1);
+	(*pv1).s2 = SKEIN_ROT(as_uint2((*pv1).s2), rc.s2);
+	(*pv1).s3 = SKEIN_ROT(as_uint2((*pv1).s3), rc.s3);
 	*pv1 ^= *pv0;
 }
 
--- a/xmrstak/backend/amd/amd_gpu/opencl/wolf-skein.cl
+++ b/xmrstak/backend/amd/amd_gpu/opencl/wolf-skein.cl
@@ -22,68 +22,59 @@ static const __constant ulong SKEIN512_256_IV[8] =
 	0xC36FBAF9393AD185UL, 0x3EEDBA1833EDFC13UL
 };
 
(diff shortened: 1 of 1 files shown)


xmrstak/backend/amd/amd_gpu/opencl/wolf-skein.cl
```diff
@@ -22,68 +23,59 @@ static const __constant ulong SKEIN512_256_IV[8] =
...
-
-void SkeinMix8(ulong4 *pv0, ulong4 *pv1, const uint rc0, const uint rc1, const uint rc2, const uint rc3)
+void SkeinMix8(ulong4 *pv0, ulong4 *pv1, const ulong4 rc)
 {
 	*pv0 += *pv1;
-	(*pv1).s0 = SKEIN_ROT(as_uint2((*pv1).s0), rc0);
-	(*pv1).s1 = SKEIN_ROT(as_uint2((*pv1).s1), rc1);
-	(*pv1).s2 = SKEIN_ROT(as_uint2((*pv1).s2), rc2);
-	(*pv1).s3 = SKEIN_ROT(as_uint2((*pv1).s3), rc3);
+	*pv1 = rotate(*pv1, (ulong4)rc);
```
taisel: I don't understand why a single rotate builtin wasn't used in the first place...

simonlui: I did some digging and it's apparently a quirk of AMD's GCN artitecture which doesn't optimize well with rotate with OpenCL which dates back to Tahiti and earlier. Not sure if that has changed now but you should do some regression testing on different hardware to see if that is still indeed the case.

See also:
https://github.com/magnumripper/JohnTheRipper/issues/1819
http://www.openwall.com/lists/john-dev/2015/10/10/1

taisel: Which is why I try to test these modifications. I had a small hash rate increase when I patched this. 

Even if rotate doesn’t compile down to a native instruction, the compiler will see a native vector operation rather than that exploded hunk of individual operations that simulate a rotate. The built in at the very least has a macro internally that may beat out what was seen earlier.

taisel: Additionally, this removes a dependency on one of the amd-only extensions. One step closer to universal opencl code.

psychocrypt:
Thank you for this pull request, I will check this PR soon. I am still waiting for my new system with AMD hardware.

mvpel:
I have been running this optimization since yesterday evening, and got positive results on an Intel processor:

```
HASHRATE REPORT - CPU
| ID |    10s |    60s |    15m | ID |    10s |    60s |    15m |
|  0 |   75.8 |   77.1 |   77.1 |  1 |   81.3 |   81.2 |   81.2 |
|  2 |   76.3 |   75.7 |   75.4 |  3 |   78.4 |   77.5 |   77.3 |
-----------------------------------------------------
Totals:    311.8  311.5  311.1 H/s
Highest:   335.3 H/s
```

Current code:

```
HASHRATE REPORT - CPU
| ID |    10s |    60s |    15m | ID |    10s |    60s |    15m |
|  0 |   75.0 |   69.2 |   67.9 |  1 |  78.2 |   73.5 |   72.5 |
|  2 |   76.9 |   72.6 |   69.5 |  3 |   76.6 |   71.4 |   69.4 |
-----------------------------------------------------
Totals:    306.7  286.7  279.3 H/s
Highest:   345.3 H/s
```

t0kieu:
@mvpel Probably this test is obsolete because CPU use different backend, not OpenCL.

caius72:
I tried this change on both RX 580 and RX VEGA64. It produces 50% invalid results. Everything went back to normal after I backed those changes out.

xmrstak/backend/amd/amd_gpu/opencl/wolf-skein.cl
mvpel: @simonlui - perhaps use function multi-versioning or a function handle to detect this quirk and apply the appropriate approach at runtime based on the GCN architecture?

psychocrypt:
I got reproduce able invalid results with this PR. Tested with a RX570 on linux.
```
1 | AMD Invalid Result GPU ID 0      | 2018-01-24 21:08:18 |
```
I will check if I can find the source of the issue.

psychocrypt:
I removed the usage of the rotate function. This was the reason for the invalid results. I will merge this PR as soon as the CI passed.

taisel:
Interesting. I’ve yet to get an invalid result with the original edits I made. I even checked the pool I use to make sure the hash rates are matching. 

I don’t use the blockchain drivers and use 460/480s. Also using a skylake cpu. In case anyone has haswell or older don’t test with my own compiled version.

taisel:
capture

psychocrypt:
I will merge it without the rotate. This has no visible impact on the speed.

PR #841, opened 2018-01-09
Fix cache size detection

tdoem:
aeon support screwed up this one. Cache size was effectively off by a factor of 1024.


psychocrypt:
Thank you for spotting. I checked your change and saw that I fully broke the auto suggestion with the commit before. Therefor I think I will fix it within one commit.
I will report later if I merge this.

psychocrypt:
I included you PR in the fix #846. Thx for you work.

PR #845, opened 2018-01-09
Increase cache threads to 10

dancudds:
Useful for cache more than 10MB

### DIFF
--- a/xmrstak/backend/cpu/crypto/cryptonight_aesni.h
+++ b/xmrstak/backend/cpu/crypto/cryptonight_aesni.h
@@ -749,7 +749,7 @@ void cryptonight_penta_hash(const void* input, size_t len, void* output, crypton
 
 // This most lovely creation will do 6 cn hashes at a time.
 template<size_t MASK, size_t ITERATIONS, size_t MEM, bool SOFT_AES, bool PREFETCH>
-void cryptonight_hexa_hash(const void* input, size_t len, void* output, cryptonight_ctx** ctx)
+void cryptonight_penta_hash(const void* input, size_t len, void* output, cryptonight_ctx** ctx)
 {
 	for (size_t i = 0; i < 6; i++)
 	{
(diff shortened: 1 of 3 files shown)


fireice-uk:
There is not enough registers in cpu to fit 6 or more interleaved hashes. You will actually make things slower by hitting the stack all the time.

dancudds:
@fireice-uk is 5 the limit despite larger cache?

fireice-uk:
Yes.

dancudds:
I seem to make it a lot faster but some invalid shares

fireice-uk:
Please post the cpuinfo, os, config and the result of `xmr-stak -V` and I will have a look.

sfalpha:
I tried 6 and it's faster on some CPU. But still with invalid shares.
7 or more got no improvement or slower.

simonlui:
@fireice-uk I'm confused. Unless you are strictly talking about x86, which has only 8 registers to use which would explain the limit, couldn't this be implemented on a conditional basis if the binary was compiled into a 64 bit binary because x86-64 has double the registers as x86 (16)?

fireice-uk:
@simonlui Registers don't work the way you think they do. It takes much more than just a single number (a register) to make a calculation, let alone a thread. In this particular case we have 16 XMM registers and 16 general registers as x86 is unsupported.

Closing as @dancudds seems to have disappeared.

dancudds:
Sorry I had left it so long - a lot of this was beyond me. @fireice-uk what do you need from me? Is it worth debugging this?

simonlui:
@fireice-uk I'm still confused, are you saying that xmr-stak is already taking advantage of x86-64's 16 registers and that's still not enough to store more than 6 hashes without spilling over CPU cache?

dancudds:
@simonlui do you have experience in 64 bit? I added code for up to 10 cache threads in my branch but I get invalid shares the more I increase it

PR #864, opened 2018-01-11
Add --needed to pacman install

Morganamilo:
Only install if the user doesnt already have the packages installed, no
point in reinstalling everything.


psychocrypt:
thx for this update

PR #882, opened 2018-01-13
fix set affinity for windows

psychocrypt:
Ignore any affinity >=64 and throw a warning.


fireice-uk:
Now, let's find a quad-cpu server to test... :D

psychocrypt:
I think nobody will run a quad socker server with windows ;-)
But fixed is fixed. We have somewhere a issue where someone run a 80 core
windows server but I can not find it anymore.

PR #977, opened 2018-01-25
avoid expired blocks

psychocrypt:
- change `pool_id` in `globalStats` to atomic
- check if the pool has changed during a bunch of hashes is calculated

Shares from the AMD and NVIDIA back-end will not be submitted if the job has changed. In the case the pool is changed in the same moment the shares will be send else the user will lose all solved shares if we switch to the dev pool. 

- [x] runtime tests AMD (8h)
- [x] runtime tests NVIDIA (2h)

### DIFF
--- a/xmrstak/backend/amd/minethd.cpp
+++ b/xmrstak/backend/amd/minethd.cpp
@@ -172,6 +172,7 @@ void minethd::consume_work()
 {
 	memcpy(&oWork, &globalStates::inst().oGlobalWork, sizeof(miner_work));
 	iJobNo++;
+	lastPool=globalStates::inst().pool_id.load(std::memory_order_relaxed);
 	globalStates::inst().iConsumeCnt++;
 
 }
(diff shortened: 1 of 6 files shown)


JoKeRz42o:
@psychocrypt Thank you for the update. Should I wait for all checks to complete or can I apply this update and test it out?

Again, thank you for all the hard work.

psychocrypt:
You can already test this PR. I did tests on linux for this PR.

fireice-uk:
I already messaged you about it, but I checked, and Snipa's pool correctly accepts previous jobs. This pr is simply throwing away good shares.

https://github.com/Snipa22/nodejs-pool/blob/1361761d5b4479f6dc6f3a86d1e7ba81f555dfc7/lib/pool.js#L762

psychocrypt:
After a long discussion with @fireice-uk  I decide to close this PR. The reason is that this PR will only avoid sending expired shares but this is only needed because some pools have not implemented that the miner get credited also for shares of the last block which are also valid.
This patch will not increase the hash rate also if the statistics looks better because of the lesser errors.

**If you have a lot of expired blocks please open an issue on you pool's website that they should implement that all shares will be credited**

MoneroOcean:
Am I right that xmr-stak miner continue to hash until it hits job target share difficulty despite the fact that pool already sent a new job for new block template with updated height?

fireice-uk:
@MoneroOcean You are wrong. The issue that we were deciding here is what to do about the work that is currently being worked on by the GPU as a new job arrives. The options are to keep on working until the batch is complete (about 1 second), or throw away the batch. 

Given that the problems are only due to primitive handling of shares by Nicehash, we elected not to change stuff.

PR #978, opened 2018-01-25
avoid nonce overlappping

psychocrypt:
The cpu miner backend uses the wrong ranges of nonces instead of using `[startNonce,startNonce + nonce_chunk)`
(startNonce,startNonce + nonce_chunk]` is used. This will results in an overlap with nonces used by the gpu back-ends.

### DIFF
--- a/xmrstak/backend/cpu/minethd.cpp
+++ b/xmrstak/backend/cpu/minethd.cpp
@@ -448,7 +448,7 @@ void minethd::work_main()
 				globalStates::inst().calc_start_nonce(result.iNonce, oWork.bNiceHash, nonce_chunk);
 			}
 
-			*piNonce = ++result.iNonce;
+			*piNonce = result.iNonce++;
 
 			hash_fun(oWork.bWorkBlob, oWork.iWorkSize, result.bResult, ctx);
 
@@ -637,7 +637,7 @@ void minethd::multiway_work_main(cn_hash_fun_multi hash_fun_multi)
 			}
 
 			for (size_t i = 0; i < N; i++)
-				*piNonce[i] = ++iNonce;
+				*piNonce[i] = iNonce++;
 
 			hash_fun_multi(bWorkBlob, oWork.iWorkSize, bHashOut, ctx);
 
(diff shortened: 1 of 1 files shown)


psychocrypt:
I changed the titel, this PR is independent of nicehash, this bug effects the whole nonce calculation.

jkuczera:
I'm now getting "Share above target." with Nicehash instead of "Invalid nonce; is miner not compatible with NiceHash?" with this PR. I'm going to wager a guess that this now only conflicts with NiceHash pool.

fireice-uk:
Your PR is generating invalid hashes.

Can you explain `The cpu miner backend uses the wrong ranges of nonces instead of using [startNonce,startNonce + nonce_chunk)
(startNonce,startNonce + nonce_chunk]` is used. `?

psychocrypt:
I used the [mathematics interval](https://en.wikipedia.org/wiki/Interval_(mathematics)) notation to to describe valid nonces which can be used based on the `start_none` from [here](https://github.com/fireice-uk/xmr-stak/blob/2ae7260b90fe3dbe835ba2489519510f0e57d770/xmrstak/backend/cpu/minethd.cpp#L437) .

e.g. if the `result.iNonce` is `20` and `nonce_chunk` is 4 than `[20,24)` are allowed nonces.
In line [440](https://github.com/fireice-uk/xmr-stak/blob/2ae7260b90fe3dbe835ba2489519510f0e57d770/xmrstak/backend/cpu/minethd.cpp#L440) `result.iNonce` is first increased and than assigned to `*piNince`, this means that the first nonce used by the miner is `21` instead of `20` in our example. The last used nonce is `24` instead of `23`.

The gpu miner uses the correct intervals for the nonce from `[startNonce,startNonce+nonce_chunk). Therefore the used nonces of the GPU and CPU will overlap by one.

This PR fixes the issue by changing `++result.iNonce` to `result.iNonce**` to first assign the old value and than increment to `result.iNonce`

psychocrypt:
I realiced that if I change the `++` I need also to handle it corecly [here](https://github.com/fireice-uk/xmr-stak/blob/2ae7260b90fe3dbe835ba2489519510f0e57d770/xmrstak/backend/cpu/minethd.cpp#L445)

psychocrypt:
OHh It looks like everything is wrong. `result.iNonce` is the next free nonce therefore this PR can be closed.

psychocrypt:
I checked everything again and the nonce intervals used are wrong and needs to be fixed

psychocrypt:
I can not reopen this PR because I forced push to the branch, therefore I opened a new PR #996

PR #995, opened 2018-01-27
Update compile instructions for AMD GPU error

minerbird:
Added instructions on how to solve compile error for AMD GPU/AMDAPPSDK. Also cleaned up external links.

### DIFF
--- a/doc/compile_Linux.md
+++ b/doc/compile_Linux.md
@@ -2,15 +2,15 @@
 
 ## Install Dependencies
 
-### AMD APP SDK 3.0 (only needed to use AMD GPUs)
+### AMD APP SDK 3.0 (only needed for AMD GPUs)
 
-- download and install the latest version from [http://developer.amd.com/amd-accelerated-parallel-processing-app-sdk/](http://developer.amd.com/amd-accelerated-parallel-processing-app-sdk/)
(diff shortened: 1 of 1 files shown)


doc/compile_Linux.md
```diff
@@ -24,6 +24,11 @@
     cmake ..
     make install
 
+    # If you get an error while compiling for an AMD GPU (e.g. make[2]: *** No rule to make target '/opt/AMDAPPSDK-3.0/lib/x86_64/libOpenCL.so',   needed by 'bin/libxmrstak_opencl_backend.so'. Stop.) execute the following and then compile again:
+
+    $ cd $AMDAPPSDKROOT/lib/x86_64
+    $ sudo ln -sf sdk/libOpenCL.so.1 libOpenCL.so
```
psychocrypt: I am not sure if this is correct. On my system it does not work until I installed the block chain driver with `--compute`. The `.so` from the sdk folder was not able to handle my gpu. ON which system do you tested this?

Spudz76:
Better to link against the .so from the active driver anyway and not the SDK version.  So this is either incorrect or non-optimal or both.

I usually `locate libOpenCL.so` and pick the one that looks like it came from the driver.

This is also why it automatically works with the blockchain driver installed, it puts the .so in the base lib64 location so it gets found first by CMake.

PR #1026, opened 2018-02-01
Fix Disabling AMD GPUs

ryanhattam:
The AMD jconf.cpp would only accept an array. The config sample, and the nvidia and cpu config work with the value 'null', as they accept a 'kNullType'. This means at current, AMD GPUs could not be turned off, the config file wouldn't parse on load.

This change makes AMD consistent with the others, and can now be disabled.


psychocrypt:
thx for you PR! well done!

PR #1062, opened 2018-02-10
Add rig identifier

fireice-uk:
Tested on snipa's pool, minergate and nanopool.

- [x] Needs to be tested on nicehash.


psychocrypt:
Is this a common tag or only supported by some pools? 
What should be tested with nicehash? I run it on the nicehash pool but worker names are only shown if the name is added to the address [documentation](https://www.nicehash.com/help/how-to-create-a-worker)

fireice-uk:
That's correct. The only thing to be tested  is whether the users are still able to log in. We can't rewrite the poolside code obviously.

psychocrypt:
Ok than I will merge this PR tomorrow. Please ignore the OSX build issue I
will fix it soon.

Simaex:
I can confirm that for xmr-stak 2.3.0 nicehash worker name is correctly identified and user can log in only if using addres string in the form "YouBTCaddress.YouWorkerName". Setting "rig_ID" parameter is not working.

PR #1087, opened 2018-02-16
AMD: option `mem_chunk`and new `strided_index`

psychocrypt:
Allow to fine tune the memory pattern of the scratchpad memory. Different AMD architectures can benefit from chunked memory blocks. Special layouts can reduce the memory load conflicts.
This PR changing not the old default but give the user the option to fine tune the device.

- add new option for `strided_index`
- add additional option if `strided_index == 2` to controll the memory chunk with

direct download: https://github.com/psychocrypt/xmr-stak/archive/topic-blockedStride3.zip

# Tested

- [x] AMD RX570 +5% hash rate

# How to test

- compile the PR
- remove your `amd.txt`
- start the miner
- edit `amd.txt` and add your old intensity and double threads per device if needed
- test the hash rate
- tweak `mem_chunk` (values between 0 and 18 are allowed)

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/gpu.cpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.cpp
@@ -332,8 +332,8 @@ size_t InitOpenCLGpu(cl_context opencl_ctx, GpuContext* ctx, const char* source_
 
 	char options[256];
 	snprintf(options, sizeof(options), 
-		"-DITERATIONS=%d -DMASK=%d -DWORKSIZE=%llu -DSTRIDED_INDEX=%d", 
-		hasIterations, threadMemMask, int_port(ctx->workSize), ctx->stridedIndex ? 1 : 0);
+		"-DITERATIONS=%d -DMASK=%d -DWORKSIZE=%llu -DSTRIDED_INDEX=%d -DMEM_CHUNK=%d",
+		hasIterations, threadMemMask, int_port(ctx->workSize), ctx->stridedIndex, int(1u<<ctx->memChunk));
 	ret = clBuildProgram(ctx->Program, 1, &ctx->DeviceID, options, NULL, NULL);
 	if(ret != CL_SUCCESS)
 	{
@@ -696,6 +696,13 @@ size_t InitOpenCL(GpuContext* ctx, size_t num_gpus, size_t platform_idx)
 
 	for(int i = 0; i < num_gpus; ++i)
 	{
+		if(ctx[i].stridedIndex == 2 && (ctx[i].rawIntensity % ctx[i].workSize) != 0)
+		{
+			size_t reduced_intensity = (ctx[i].rawIntensity / ctx[i].workSize) * ctx[i].workSize;
+			ctx[i].rawIntensity = reduced_intensity;
+			printer::inst()->print_msg(L0, "WARNING AMD: gpu %d intensity is not a multiple of 'worksize', auto reduce intensity to %d", ctx[i].deviceIdx, int(reduced_intensity));
+		}
+
 		if((ret = InitOpenCLGpu(opencl_ctx, &ctx[i], source_code.c_str())) != ERR_SUCCESS)
 		{
 			return ret;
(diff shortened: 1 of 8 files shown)


Nathilion:
I compiled the PR using Visual Studio 2017 v15.5.6 and its Cmake plugin. I'm running this on Windows 10 Home v1709, build 16299.248. AMD driver version 18.2.1.

The test results are for a Sapphire R9 280x. Stock clocks of 1020MHz GPU, 1500MHz memory.

xmr-stak 2.2.0 has a hash rate of ~440 H/s with the following config:
"gpu_threads_conf" : [
  // gpu: Tahiti memory:2265
  // compute units: 32
  { "index" : 0, "intensity" : 896, "worksize" : 8, "affine_to_cpu" : false, "strided_index" : true },
],

Test executable:
With the same settings I get the same has hash rate.

Double thread test, varying intensity. Default generated config otherwise:
"gpu_threads_conf" : [
  // gpu: Tahiti memory:2265
  // compute units: 32
  { "index" : 0,
    "intensity" : 448, "worksize" : 8,
    "affine_to_cpu" : false, "strided_index" : 2, "mem_chunk" : 4
  },
  { "index" : 0,
    "intensity" : 448, "worksize" : 8,
    "affine_to_cpu" : false, "strided_index" : 2, "mem_chunk" : 4
  },
]

I=intensity
I=448: 210+210
I=560: stable. 210+210
I=616: stable 186+185 but hash rate varies quite a bit from 160 - 200.
I=672: driver crash
I=728: driver crash
I=768: driver crash
I=896: stable but with tons of _Error CL_MEM_OBJECT_ALLOCATION_FAILURE when calling clEnqueueNDRangeKernel for kernel 0._ Hash rate is 418+(300-627). I wonder if the rate calculation works properly in this situation. Getting 1000 hashes per second would be awesome despite all the allocation failures.

Mem chunk variations:
1 thread   I=896, mem 0: 430 H/s
1 thread   I=896, mem 4: 428
1 thread   I=896, mem 18: 429
2 threads: I=560, mem 0: 210+210
2 threads: I=560, mem 1: 210+210 but a crash after a minute or two.
2 threads: I=560, mem 2: 210+212
2 threads: I=560, mem 3: 210+212
2 threads: I=560, mem 8: 190+190. After a few minutes got up to 202+209.
2 threads: I=560, mem 18: 212+210

I have a few questions though: how is this strided memory supposed to work? What is the idea behind? It feels like black magic. Although that can be said in general about figuring out what works best for any given graphics card, haha.

JerichoJones:
AMD Vega 64 Reference Card  
image

psychocrypt:
@Nathilion By changing the layout of the scratchpad memory the conflicting memory loads can be reduced. Normally all threads in a work group  should read contiguous memory. Due to the the random memory access within a scratchpad for each thread this is not possible.

raphaelans:
## xmr-stak  v2.2.0

RX 580 8GB - 2 threads
"intensity" : 1152, "worksize" : 8,
"affine_to_cpu" : false, "strided_index" : true

Vega 56 Strix - 2 threads
 "intensity" : 1932, "worksize" : 8,
 "affine_to_cpu" : false, "strided_index" : true

xmr_stak_old_version

## xmr-stak  v2.2.0-737185ee

report_2_auto_config

report_2

report_3

report_4

report_5

report_6

psychocrypt:
@raphaelans Thx for your results. I am confused by your test results. You
set strided_index to 1 and changed mem_chunk. mem_chunk has only a effect
if strided_index is 2.
Is there a mistake in the configs pasted unto the screen shots?

raphaelans:
@psychocrypt the 1st image to the 5th image I make comparison from JerichoJones's test result and the 6th image (the last one image) I use stridex_index = 2 for first thread each gpu and for the second thread I don't use stridex_index = 1 and forgot to change mem_chuk = 0.

Do you need I test strided_index enable (2) for single and double threads for each gpu for mem_chuck result?

psychocrypt:
Thanks for the offer to make additional tests. I think the current tests
are  enough.
It looks like the new strided_index give only in some cases a benefit. It
could be that it depends on the driver.
Never the less the community will find the best settings after the release
:-)

raphaelans:
@psychocrypt you're welcome :) also many thanks for your hardwork and for all who have contributed this great project.

psychocrypt:
For all who tested this pull request I implemented a bug thus mem_chunk was not `pow(2,mem_chunk)` instead `mem_chunk` was really the number of elements. This is now fixed in #1109.
There is no need to test it but it could be that the fixed version can give an advantage for higher `mem_chunk` values.

PR #1097, opened 2018-02-19
Update donate-level.hpp

IvarSandin:


No0n3Left:
I do partially get the developer donation, though having it compiled in is just a hastle. I would be more inclined to leave the donation enabled if there were a config option modifiable at runtime instead of a compiled in option.

Issue #1018, opened 2018-01-31
Improve Device Initialisation Start Up Time


guerrillatiger:
Hi,

This issue is a duplicate of #201 however there was no resolution to it and it was closed.

Issue 
During initial startup xmr-stak is 'Compiling code and initialising GPUs', this takes about 5 seconds per thread. Is it possible to reduce this? When I want to change the coin I am mining I would like start up to be as quick as possible, this is approximately 1.5min.

Possible Solution 1:
Compile threads in parallel as my CPU only gets to make 30% during startup.

Possible Solution 2:
If no changes are made in the gpu configuration is there a need to compile again? Could this compiling be cached between starts, perhaps pass a value to say use cache if it exists.

Reproduction:
Start miner.

Rig Setup
OS: Windows 10 Pro 1709 16299.192
XMR Version: xmr-stak/2.2.0/c4400d19/master/win/nvidia-amd-cpu/aeon-monero/20
CPU: Intel Core i3-7100 (no CPU mining running)
GPU: AMD Vega 64 x8, undervolted and overclocked. Runs stable.

![xmr-stak slow start](https://user-images.githubusercontent.com/35990637/35612624-76646262-067a-11e8-8eeb-834dc8bd5f5c.PNG)

![xmr-stak cpu usage](https://user-images.githubusercontent.com/35990637/35612629-7ae2b9f6-067a-11e8-9501-41729f8fc5d4.PNG)

[amd.txt](https://github.com/fireice-uk/xmr-stak/files/1680905/amd.txt)

[clinfo_output.txt](https://github.com/fireice-uk/xmr-stak/files/1680907/clinfo_output.txt)

silviotech:
I too am interested in this, i have a vega 64 FE, and created a little script to restart the program / disable and re-enable video card if necessary. Although i have quite stable settings currently sometimes it restarts numerous times. I thought it would be cool to have an option to pre- compile those settings for our specific AMD setup.

khobbs19:
How about if config.txt has changed or keyboard input eg ‘x’ to re-read config.txt (not amd.txt etc) without restarting the miner? Thanks!

eliashdezr:
Yep, hash the config files and if they don't change on each start, recompile for each GPU, this would decrease a lot the start time.

jam3972:
[...]

PR #1131, opened 2018-03-03
OpenCL compile cache

psychocrypt:
Reduce OpenCL start time by using a self made compiler cache.

- store compiled OpenCL binary
- load OpenCl binary if available
- add project picosha2

solve: #1018
# How it works

- the compiled binary is stored into a file during the first start
- if the code and compile parameter not change the binary is loaded with the second start of the miner

# Review

ignore white spaces during the code review [direct link to the diff without white space changes](https://github.com/fireice-uk/xmr-stak/pull/1131/files?w=1)

# Tests
- [x] WIN64 AMD test
- [x] linux test
- [x] need more tests

### DIFF
--- a/THIRD-PARTY-LICENSES
+++ b/THIRD-PARTY-LICENSES
@@ -22,3 +22,8 @@ License: MIT License and BSD License
 
 -------------------------------------------------------------------------
 
+Package: PicoSHA2
+Authors: okdshin
+License: MIT License
+
+-------------------------------------------------------------------------
--- a/xmrstak/backend/amd/amd_gpu/gpu.cpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.cpp
@@ -15,6 +15,7 @@
 
 #include "xmrstak/backend/cryptonight.hpp"
 #include "xmrstak/jconf.hpp"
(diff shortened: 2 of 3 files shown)


Norfolc:
I've checked your fixes. It works great!
For my configuration (6 identical Vega56) it runs compilation only for the first thread and then uses cache for other threads (for the first run). For other runs it just uses the cache and starts up in 2 seconds.
As a result startup time for the first time decreased from 165 to 16 seconds.

omonji:
Same for me this is awesome!!!

jam3972:
Seeing similar results! This is perfect! Vega owners rejoice!
Thank you!
Stall on fresh Vega FE install runs still happen, but I'll open separate issue for that.

GabrielKesler:
Merged this PR into dev and it works perfectly on a 7 x Vega64 rig.

Cheers!

psychocrypt:
@fireice-uk This is now ready to merge

caius72:
This works great! Thanks @psychocrypt for the enhancement. I too was experimenting with threads before and also noticed that it only yielded a 10-15% speed up.

PR #1137, opened 2018-03-04
Add prometheus monitoring support.

yorik:
Add /metrics page with following metrics:
    connection_uptime
    difficulty
    good_results
    total_results
    threads
    hashes_count


t0kieu:
Hi, that implementation is not very good :-/
1. Prometheus has own library for metrics also ccp implementation.
2. Your names of metric keys doesn't use Prometheus conventions.
3. Some keys are redundant by convention (Prometheus practices) should be used labels.

I have implemented prometheus metrics too, I will do PR in near future. I need only rebase code.

yorik:
RE 1: I thought that will be one more dependency while it's quite easy to implement from scratch.
      2: It was a quick hack I made for my own purposes, so proper names would make sense.
      3: Agree, also having errors would be more useful.

Could you show me your PR?

t0kieu:
Hi, I will try focus on that today.

PR #1161, opened 2018-03-15
reenable MACOSX travis tests

psychocrypt:
- remove `brew trap science`

MACOSX travis tests was removed with #1066 to pass the CI.

- [ ] do not merge before the travis tests passed


psychocrypt:
this was only a try if travis can be reanabled if I opened a new PR

PR #1194, opened 2018-03-25
Fixing allocation issue / CUDA

jleni:
When initializing CUDA context, there is no enough space for bigger blobs.
Increasing context allocation from 84 to 112 bytes.


psychocrypt:
thx to solve the merge conflicts I included this PR in #1198

fireice-uk:
@jleni The protocol limit (practical) is 84 bytes, unless versions are more than 127 or it is 2030 (timestamp)

PR #1195, opened 2018-03-25
instructions for new gcc versions

jleni:
Adding additional instructions for Linux distributions with new gcc compilers.


psychocrypt:
thx for this update

PR #1196, opened 2018-03-25
Splitting types/functions to simplify wrapper creation

jleni:
This allows for exposing the types independently from the function declarations.
This is necessary to embed/mock xmr-stak in other projects and improves testability.


fireice-uk:
I might be slightly edgy because of overwork, but I don't see the benefit of that change. Allocating a context is a fairly complex job (especially LP one). And right now they are both in the same header. What's the point of splitting them up?

PR #1230, opened 2018-03-27
Patch 2

Simaex:


psychocrypt:
thy for your pull request. Each pool has a different way how a miner name can be set. This is also independent  of the OS. Therefore I opened #1238 instead of this PR.

PR #1234, opened 2018-03-27
run xmr-stak without admin privileges

zifootball:
i want to run xmr-stak without admin privileges please help me


psychocrypt:
with `xmr-stak --help` you can see all available option. one is `--noUAC                    disable the UAC dialog`

PR #1236, opened 2018-03-27
POW AEON v7

psychocrypt:
- add new pow for AEON
- fix missing cryptonight-heavy selection for multi hashes

# Tests

- [ ] AMD, CPU (low_power 1,2,3,4,5), AMD
```
"pool_list" :
[
{"pool_address" : "aeonrebase.supportcryptonight.com:3333" , "rig_id" : "", "wallet_address" : "UmiVUjz1wsWf8QsZ9dnB96FqJ1pPruT13P7RgaaHzAzeePW7r4oP3PhhViByBhFzivVSF7uNVtsqLeaYvXcwyY9G1hRbAhk4X", "pool_password" : "x", "use_nicehash" : false, "use_tls" : false, "tls_fingerprint" : "", "pool_weight" : 10 },
],

"currency" : "aeon7"
```

### DIFF
--- a/xmrstak/misc/executor.cpp
+++ b/xmrstak/misc/executor.cpp
@@ -571,6 +571,13 @@ void executor::ex_main()
 			pools.emplace_front(0, "donate.xmr-stak.net:4444", "", "", "", 0.0, true, false, "", true);
 		break;
 
+	case cryptonight_aeon:
+		if(dev_tls)
+			pools.emplace_front(0, "donate.xmr-stak.net:7777", "", "", "", 0.0, true, true, "", true);
+		else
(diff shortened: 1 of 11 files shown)


xmrstak/misc/executor.cpp
```diff
@@ -571,6 +571,13 @@ void executor::ex_main()
 			pools.emplace_front(0, "donate.xmr-stak.net:4444", "", "", "", 0.0, true, false, "", true);
 		break;
 
+	case cryptonight_aeon:
```
fireice-uk: This isn't needed. In cases like this it is much better to do:

`case cryptonight_lite:`
`case cryptonight_aeon:`
`    [...]`
`    break;`

Can you correct the all switch statements in your to this format?

PR #1237, opened 2018-03-27
Fixed typo in xmrstak/backend/amd/amd_gpu/gpu.cpp

djfinch:
cod --> code

PR #1240, opened 2018-03-27
fix parentheses warning

psychocrypt:
fix warning

```
/Users/user/xmr-stak/xmrstak/backend/amd/amd_gpu/gpu.cpp:481:11: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
                if( ret = clGetProgramInfo(ctx->Program, CL_PROGRAM_BINARIES, num_devices * sizeof(char*), all_programs.data(),NULL) != CL_SUCCESS)
```


fireice-uk:
Ooops, looking at #1237 I owe @djfinch an apology for typing nonsense in his pr =).

djfinch:
Not needed but thanks anyways :). Thanks for your awesome work btw. :) 👍

PR #1284, opened 2018-04-03
AMD OpenCL: fix sumokoin

psychocrypt:
- fix that version argument was not passed to extended kernel parameters


psychocrypt:
PR against master, maybe it is to early^^

PR #1299, opened 2018-04-04
fix cuda architecture detection

psychocrypt:
fix #1297

If sm_20 is mixed with other architectures the detection for the minimal supported architecture is broken.

PR #1301, opened 2018-04-04
Bug fix release 2.4.2

psychocrypt:
Bug fix release 2.4.2
- fix sumokoin for CUDA #1300
  - fix sm_20
  - fix wrong results if `bfactor` is used
- fix wrong CUDA architecture detection #1299


nightsuspect:
this means at 2.4.2 nvidia's auto config will return back that removed with 2.3.0 ?

psychocrypt:
@nightsuspect no this is a bugfix release because sumokoin is broken

PR #1339, opened 2018-04-06
Use EXECUTABLE_OUTPUT_PATH and LIBRARY_OUTPUT_PATH

candrews:
Use EXECUTABLE_OUTPUT_PATH and LIBRARY_OUTPUT_PATH to control the path to which the executable and libraries are installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -522,8 +522,8 @@ else()
     add_executable(xmr-stak ${SRCFILES_CPP})
 endif()
 
-set(EXECUTABLE_OUTPUT_PATH "bin")
-set(LIBRARY_OUTPUT_PATH "bin")
+set(EXECUTABLE_OUTPUT_PATH "bin" CACHE STRING "Path to place executables relative to ${CMAKE_INSTALL_PREFIX}")
+set(LIBRARY_OUTPUT_PATH "bin" CACHE STRING "Path to place libraries relative to ${CMAKE_INSTALL_PREFIX}")
 
 target_link_libraries(xmr-stak ${LIBS} xmr-stak-c xmr-stak-backend)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -535,23 +535,23 @@ target_link_libraries(xmr-stak ${LIBS} xmr-stak-c xmr-stak-backend)
 # do not install the binary if the project and install are equal
 if( NOT CMAKE_INSTALL_PREFIX STREQUAL PROJECT_BINARY_DIR )
     install(TARGETS xmr-stak
-            RUNTIME DESTINATION "${CMAKE_INSTALL_PREFIX}/bin")
+            RUNTIME DESTINATION "${CMAKE_INSTALL_PREFIX}/${EXECUTABLE_OUTPUT_PATH}")
     if(CUDA_FOUND)
         if(WIN32)
             install(TARGETS xmrstak_cuda_backend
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -522,8 +522,8 @@ else()
     add_executable(xmr-stak ${SRCFILES_CPP})
 endif()
 
-set(EXECUTABLE_OUTPUT_PATH "bin")
-set(LIBRARY_OUTPUT_PATH "bin")
+set(EXECUTABLE_OUTPUT_PATH "bin" CACHE STRING "Path to place executables")
```
psychocrypt: please add extent the description that this path is relative to `${CMAKE_INSTALL_PREFIX}`

candrews: I forced pushed the updated working

candrews:
@psychocrypt so think this can be merged now?

Spudz76:
Honestly I don't think the backend libs belong in a system-libs directory.
They are not usable by anything else like things that DO belong there.
When a lib is only for a specific executable, they belong near that executable.
Also I'm not sure I would install xmr-stak as a global/system utility either.  It's designed to run in its own homedir/sandbox/docker hole.
It should be more like the Gentoo packages that have "instances" to them (web apps, etc) and live in a jail by themselves.

candrews:
I agree that the libraries should be placed in a private directory (I think /usr/lib/xmr-stack would be great), but it's also not a big deal that they're in /usr/lib. Libraries do belong in lib, not in bin - having libraties "near" the executable isn't how Linux library packaging works.
You're welcome to run this or anything else however you like (in docker, flatpak, whatever), but for Gentoo, we comply with the Gentoo standards. If you have a problem with the Gentoo packaging, I'm happy to address it - please report a bug on the Gentoo tracker. I don't think this is the appropriate place to report issues with the Gentoo packaging.
Trying to stay on topic, I suggest this PR be merged and other issues be handled elsewhere.

Spudz76:
Shared libraries do belong in /usr/lib, agreed.

These libraries are ".so" but they are not shared, they should be protected from even being found by ldconfig in normal system operation (thus putting them in the working dir with the executable, it's a scope thing).

The executable does not belong in the system bin dirs either.  It generates config files in whatever PWD you run it from so that could get very messy (oops I typed xmr-stak, now there are a bunch of .txt files in this directory wherever I am)

However, *IF* this were actually designed for Linux, then it would use `/etc/xmr-stak/` for config files, and probably `.a` the backend libs and link them in explicit instead of leaving them "windows style" as DLL's (happen to be .so because Linux, and the separate dynamic lib thing is *REQUIRED* over there, and from there it's a consistency thing).

But, since it's essentially a Windows app that happens to compile on UNIXes coercing it into what's right for Linux is the wrong tact.  I run `/opt/xmr-stak/` and `/opt/xmr-stak-some-other-coin-this-week/` and who knows how many other "installations" of it, thus why I requested it be more like a sandbox/app deal much like you can generate multiple individual webroots with many of the webapps.  And all related files are in that one directory to force/sandbox the scope.

It reeeeealy doesn't belong hooked into the system.

You know people worm it around right, so you'd be making it even easier, say a Gentoo box on the net gets intruded, owner had installed Gentoo-ized System Level xmr-stak, all the hacker has to do is drop some configs and xmr-stak is already in the path.  How easy!  Didn't even have to go download an entire binary trojan tarball or anything, how nice of the Gentoo packagers to make this available to any old idiot that gets in.

PR #1375, opened 2018-04-08
expose original algorithm

psychocrypt:
- expose cryptonight_lite_v7, cryptonight_heavy and cryptonight_v7
- remove cryptonight_lite
- update documentation in `pools.tpl`

- [x]  ~~do not merge before #1372 to be sure that `cryptonight_heavy` can be used~~


BigslimVdub:
so removing Cryptonight_lite will still allow for current pre-fork usage using Cryptonight_lite_v7 until fork then it will auto switch?

fireice-uk:
@BigslimVdub if you are confused by the meaning of the options you should use your coin's name.

psychocrypt:
@fireice-uk  I updated the PR, I re-ordered the algorthms and add a note about the scratchpad size in the file `pools.tpl`

PR #1391, opened 2018-04-10
Add appveyor-build-artifacts

jakibaki:
This will make it possible for users/devs to download the windows-dev-builds of xmr-stak directly from appveyor.

The artifacts will be available at https://ci.appveyor.com/project/fireice-uk/xmr-stak/build/artifacts


psychocrypt:
Thanks for this nice PR, but we need to sign the binaries somehow.

There are two additional points which must be solved before:
- I need to include AMD opencl into the build
- the Ci must compile for all CUDA compute capabilities

PR #1401, opened 2018-04-11
Update donate-level.hpp

salehoptany:

### DIFF
--- a/xmrstak/donate-level.hpp
+++ b/xmrstak/donate-level.hpp
@@ -15,4 +15,4 @@
  *
  */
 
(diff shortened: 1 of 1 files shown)


Raindeer44:
You can certainly change the level to whatever you want on your own machine, but I think 2% is a reasonable default.

psychocrypt:
why not changing it to `10.0` ;-)

psychocrypt:
No thats than mean we roll the dice and the possibility that we mine to the
user pool is 5% ;-)

PR #1419, opened 2018-04-13
check plugin version

psychocrypt:
- move define version number and name of xmr-stak to `version.hpp`
- add for gpu backends the interface function `xmrstak_version_backend`
- handle wrong version of miner and backends

This PR  avoid the issue that a user only copies the binary of the miner and missed to updat ethe dynamic libraries.

# Tests

- [ ] AMD backend Windows/Linux
- [ ] NVIDIA backend Windows/Linux


psychocrypt:
I added WIP to this PR because I need to test it a little bit more if the miner still runs fine if all dll's are removed.

PR #1437, opened 2018-04-16
fix wrong algo selection

psychocrypt:
In the case where the dev pool mines on a higher version than a monero fork coin the miner is not resetting the algorithm.

- select the correct algorithm each time the block version or pool ID changed.
- remove fork version for sumokoin and monero7
- allow the dev pool to fork on a different block version than the user descriped coin

All algorithm are centered around the user coin description.
It is allowed to have two two different coin algorithms in the user coin description.
It is only allowed to use algorithms for the dev pool coin description those are used in the user coin description.
There are two ways to define a non forking coin.
  - set both user coin algorithm descriptions to the same algorithm and set version to zero
  - set the first algorithm in the user coin description to something you like to use in the dev pool
    and set the second algorithm to the correct representation of the coin. Set the version to 255.
    This will allow that the dev pool can mine on a different coin algorithm than the not forking user coin.
    Do not use an algorithm with different scratchpad size for the dev pool.

- [x] need rebase against #1375

# Tests

- [x] validated different coins before and after the fork


xmrstak/backend/amd/amd_gpu/gpu.cpp
```diff
@@ -1087,7 +1087,7 @@ size_t XMRSetJob(GpuContext* ctx, uint8_t* input, size_t input_len, uint64_t tar
 size_t XMRRunJob(GpuContext* ctx, cl_uint* HashOutput, xmrstak_algo miner_algo)
 {
 	// switch to the kernel storage
-	int kernel_storage = miner_algo == ::jconf::inst()->GetMiningAlgo() ? 0 : 1;
+	int kernel_storage = ::jconf::inst()->GetCurrentCoinSelection().GetDescription(1).GetMiningAlgo() ? 0 : 1;
```
psychocrypt: BUG!!!

fireice-uk:
#1375 has been merged, please rebase.

PR #1446, opened 2018-04-17
Add benchwait and benchwork

Spudz76:
Add options `--benchwait` and `--benchwork` to tune the waiting and working of the benchmark

Supplying those without also supplying `--benchmark` is left as an exercise in futility and end-user intelligence test (does nothing)


Spudz76:
Recreated properly as pull #1473

PR #1448, opened 2018-04-17
CUDA 8 deprecation warning silencer

Spudz76:
Have had this in my tree for a while just so the compile output is less busy with useless warnings.
I think maybe everyone knows arch 2.x went away.


psychocrypt:
I think it is not good to suppress this warning. The warning will only be shown if the compute architecture 20 is selected.
I will sleep one night over it.

PR #1451, opened 2018-04-17
Update ITNS with v4 fork to CN v1

Venthos:
Official Announcement: https://intensecoin.com/2018/04/16/imminent-hard-fork-and-network-attack/

Sorry, I'm dumb as a brick and accidentally closed the other pull request.  Sigh.

Note: Intense Coin already exists in README.md w/link to the main site and already mentioned in xmrstak/pools.tpl.  This simply updates the fork block version to Cryptonight v1 as v4 for Intense Coin.  The pool example is the official pool, which oddly does not have a host pointed to it (hence the bare IP).


psychocrypt:
I removed the IP address because the pool webpage is not available and I thin in the case the owner moved the system behind to an other hoster the IP will not stay the same.

PR #1464, opened 2018-04-19
Adding ITALOCOIN as mining option

italocoin-project:
Adding ITALOCOIN as mining option


fireice-uk:
Please note that your coin does not appear to meet our requirement of sustained 1MH/s hashrate.

italocoin-project:
@fireice-uk we are about to hit the market, the hashRate will blow up, i will squash in a moment!

Issue #1438, opened 2018-04-16
AMD OSX 10.11.6 issues Trace/BPT trap: 5


scottwils:
Please provide as much as possible information to reproduce the issue.

# Basic information
  -I get an trap error when running this.

bash-3.2$ ./xmr-stak
-------------------------------------------------------------------
xmr-stak 2.4.2 e10e8e6

Brought to you by fireice_uk and psychocrypt under GPLv3.
Based on CPU mining code by wolf9466 (heavily optimized by fireice_uk).
Based on OpenCL mining code by wolf9466.

Configurable dev donation level is set to 2.0%

You can use following keys to display reports:
'h' - hashrate
'r' - results
'c' - connection
-------------------------------------------------------------------
[2018-04-16 14:14:55] : Mining coin: monero7
[2018-04-16 14:14:55] : Compiling code and initializing GPUs. This will take a while...
[2018-04-16 14:14:55] : WARNING: using non AMD device: Apple
[2018-04-16 14:14:55] : Device 0 work size 8 / 32.
[2018-04-16 14:14:55] : OpenCL device 0 - Precompiled code /Users/t1admin/.openclcache/494bff8065cf9dffd42c862e8dbc3d015f03695fb5dbbbc858411ea566a39fa1.openclbin not found. Compiling ...
Trace/BPT trap: 5
bash-3.2$

  - Type of the GPU (if you try to miner with the GPU).
  // gpu: AMD Radeon HD - FirePro D700 Compute Engine memory:1408

# Compile issues
  - Which OS do you use?
  10.11.6
bash-3.2$ cmake -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl -DCUDA_ENABLE=OFF -DCPU_ENABLE=OFF .
-- The C compiler identification is AppleClang 8.0.0.8000042
-- The CXX compiler identification is AppleClang 8.0.0.8000042
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
[...]

PR #1486, opened 2018-04-22
fix that cli option `--noAMDCache`

psychocrypt:
- fix that to much OpenCL calls whre used if cache is disabled

This fix allow to workaround the broken MAC OSX OpenCl #1438.


scottwils:
How do I test this version on my AMD Mac with affected OS?

Spudz76:
Follow normal directions but after cloning repo and in the base project directory, do `git checkout dev` and then pump the contents of `https://github.com/fireice-uk/xmr-stak/commit/470c1cd533f7354ab72187d422072b1caf5f489f.patch` into `patch -p1`

`wget -qO- https://github.com/fireice-uk/xmr-stak/commit/470c1cd533f7354ab72187d422072b1caf5f489f.patch | patch -p1`

PR #1488, opened 2018-04-22
Remove redundant pool connection msg code

ghost:


xmrstak/misc/executor.cpp
```diff
@@ -321,11 +318,7 @@ jpsock* executor::pick_pool_by_id(size_t pool_id)
 void executor::on_sock_ready(size_t pool_id)
 {
 	jpsock* pool = pick_pool_by_id(pool_id);
-
-	if(pool->is_dev_pool())
-		printer::inst()->print_msg(L1, "Dev pool connected. Logging in...");
-	else
-		printer::inst()->print_msg(L1, "Pool %s connected. Logging in...", pool->get_pool_addr());
+	printer::inst()->print_msg(L1, "%s connected. Logging in...", pool->is_dev_pool() ? "Dev pool" : pool->get_pool_addr());
```
fireice-uk: I would prefer to have "Pool pool.example.com:3333 connected." rather than just "pool.example.com:3333"

PR #1520, opened 2018-04-30
Updated compile instructions due to AMD APP SDK no longer available on AMD website

cialu:


Spudz76:
Downvote since that's my webspace and it's not permanent

cialu:
@Spudz76 the commit shows only the link to GitHub issue. In the meantime, we can study a different solution and then put it in the issue. Maybe, a public cloud.

What do you think about?

Spudz76:
Assume AMD would eventually put it back up / their web is always like this

cialu:
It's a temporary server problem, I have updated the issue.

Issue #1337, opened 2018-04-06
NH rejects a lot of shares


kranimantas:
When I start the miner, it works well for about an hour, and then the hell breaks loose: I keep getting different errors: from "Share not found" to "Share above target". And these aren't one offs, the errors come in batches, and eventually almost all of the shares become bad.

![nh](https://user-images.githubusercontent.com/38140525/38436358-41fc2a04-3989-11e8-8ac9-ed9841b6c445.png)

With 2.2.0 there were no issues, and now the configuration is exactly the same, but I'm getting a ton of rejections. No changes in hardware, temperatures normal, no other errors or suspicious symptoms.

After a restart, everything works well for a while again, and then the rejections start happening. CAST works fine, comparable to 2.2.0 (some errors, but less than 3 percent of shares are bad).

Could this be some kind of a loop / memory leak problem that makes _all_ the shares bad? After a restart of the miner, it works well for a while again.

# Basic information
  - Type of the CPU.
AMD Ryzen 5 2400G
  - Type of the GPU (if you try to miner with the GPU).
Sapphire Vega 64

# Issue with the execution
  - Do you compiled the miner by our own?
No
  run `./xmr-stak --version-long` and add the output here
xmr-stak/2.4.2/e10e8e67/master/win/nvidia-amd-cpu/aeon-cryptonight-monero/20

# Stability issue
  - Is the CPU or GPU overclocked?
CPU no, GPU as per vega.mininguides.com

  - Is the Main memory of the CPU or GPU undervolted?
CPU no, GPU as per vega.mininguides.com

psychocrypt:
monero hash forked therefore please update to the latest miner if you try
to mine monero.
https://github.com/fireice-uk/xmr-stak/releases

kranimantas:
This is the 2.4.2 misbehaving, unfortunately.

perestoronin:
and for me, miners and problems with errors as above
`# xmr-stak -V
Version: xmr-stak/2.4.2/`

psychocrypt:
please post the config files

hchan123:
I have the same issue when mining against the cryponightv7 pool on nicehash
using 2.4.2.  The miner would mine fine for about an hour and then I will
start getting above target error for results after.

psychocrypt:
please also post the error output (press key 'r') wee need to know the
error type

perestoronin:
`gpu_threads_conf" : [
  // gpu: gfx900 memory:8048
[...]

Issue #1222, opened 2018-03-26
"Share above target" version 2.3.0 monero-7 (Nicehash)


DrStein99:
(on "MONERO7" setting)

I compiled version 2.3.0, and run on 2 of my rigs.  Runs flawless on monero pools.  I test one rig to Nicehash.  Immediately starts rejecting shares "Share above target".  

I restart the rig, try again - and the errors come right in.  I try on 1 more additional rig test and results are the same.

After 1 hour hashing both rigs, so far it's reporting 91% and 93% (but I am not sure since Nicehash does "socket closed" and the STATS become offset by it's 90 second break-time to mine another pool between Nicehash socket-closings).

I guess they have not switched to Monero7, or some other reason - I do not know.

Sandown:
Same issue here, reports miner errors within stak. 

for example:

[stratum+tcp://cryptonightv7.eu.nicehash.com:3363] PARSE error: Unsupported server method mining.set_extranonce
--

DrStein99:
I was connecting to cryptonight stratum server.  I did not use cryptonightv7.eu.nicehash.com.  But now that I know that it exists, I will try again, and use the cryptonight7 setting to re-test.

Sigura:
looks like a same as I got for 2.3.0 monero7:

[...]

PR #1526, opened 2018-05-01
fix job consume

### LINKED ISSUE #1505: Xmr-Stak2.4.3 continues mining "rejected shares" for hours after reconnecting to pool
Please provide as much as possible information to reproduce the issue.

# Basic information
V2.4.3 will run for a decent amount of time (days)  w/o any issues..... 100% accepted share rate... then immediately following an internet caused loss of connection to pool the v2.4.3 will attempt to reconnect to pool.  The reconnection appears in the log.txt as normal except ALL following shares are now "rejected shares".  The miner is NOT banned and there were NO duplicate shares nor bad shares submitted prior to the disconnection from pool (actually up to 100% accepted prior to disconnection).  It seems the reconnect is not complete as the pool shows NO shares submitted during this time.  Unfortunately, the miner continues mining the "rejected shares" for hours...... until time that it is manually turned off and restarted.
connection rejection

Intel Pentium G3250 3.2GHz   - Type of the CPU.  
Sapphire Nitro rx580 x6  - Type of the GPU (if you try to miner with the GPU).

# Compile issues
  - Which OS do you use? Windows 10 1079
  ```
  add **all** commands you used and the **full** compile output here
  ```
  ```
  run `cmake -LA .` in the build folder and add the output here
  ```

# Issue with the execution
  - Do you compiled the miner by our own?
  ```yes...................
"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" x64 -vcvars_ver=14.11

"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\Tools\VsMSBuildCmd.bat"

cd /

cd xmr-stak-2.4.3

mkdir build

cd build

set CMAKE_PREFIX_PATH=C:\xmr-stak-dep\hwloc;C:\xmr-stak-dep\libmicrohttpd;C:\xmr-stak-dep\openssl;

cmake -G "Visual Studio 15 2017 Win64" -T v141,host=x64 ..

cmake --build . --config Release --target install


    85 Warning(s)
    0 Error(s)
   Time Elapsed 00:13:00.89

cd bin\Release

copy C:\xmr-stak-dep\openssl\bin\* .


 


 run `./xmr-stak --version-long` and add the output here
  ```C:\xmr-stak-2.4.3_\build\bin\Release_x64>xmr-stak --version-long
Version: xmr-stak/2.4.3/26a5d65f/unknown/win/nvidia-amd-cpu/aeon-cryptonight-monero/0
Press any key to exit.

# AMD OpenCl issue

  ```
  run `clinfo` and add the output here
  ```

# Stability issue
  - Is the CPU or GPU overclocked?
  - Is the Main memory of the CPU or GPU undervolted?

psychocrypt:
~~fix #1505~~
~~fix or reduce #1337 #1222~~

- remove switch_work in all classes `minethd`
- move `consume_work` into `globalStates`
- remove the limit that each worker thread must be consume a job before we can receive a new job. 
- miner should not freeze if one GPU is hanging

In the previous implementation it was possible that a worker thread consumed and already old job if the pool is sending jobs faster than a batch of nonces is calculated. The reson is that the current implementation force that each worker must consume the oldest job in the queue before a new job can be added. On a GPU the time to process a job (a batch of nonces) before checking again if the job has changed much longer than on a CPU. A GPU needs typical ~1 sec+ (slow GPUs sometimes 10sec).

example of the PR behavior if the job is faster changed than processed:

- 1. job switch
- 2. thread consume
- 3. job switch 
- 4. job switch again
- 5. thread is finishing II. and never consume III

# Test

- [x] CPU
- [x] AMD
- [x] NVIDIA

[update:] I thought there is a race condition and deadlock in the code I changes. After discussions with @fireice-uk it looks like I was wrong. Therefore I updated the PR description.

### DIFF
--- a/THIRD-PARTY-LICENSES
+++ b/THIRD-PARTY-LICENSES
@@ -5,7 +5,7 @@ This application bundles the following third-party software in accordance with t
 Package: Original NVidia mining code
 Authors: tsiv and KlausT
 License: GNU GPLv3
-Notes: Improvements are (c) of Xmr-Stak team
+Notes: Improvements are (c) of Xmr-Stak team team and are covered by GNU GPLv3
 
 -------------------------------------------------------------------------
 
@@ -24,6 +24,13 @@ License: MIT License and BSD License
 
 Package: PicoSHA2
 Authors: okdshin
+License: MIT License
+
+-------------------------------------------------------------------------
+
+Package: cpputil
+Authors: Will Zhang
+Source: https://github.com/willzhang4a58/cpputil
 License: MIT License
 
 -------------------------------------------------------------------------
--- a/THIRD-PARTY-LICENSES
+++ b/THIRD-PARTY-LICENSES
@@ -5,7 +5,7 @@ This application bundles the following third-party software in accordance with t
 Package: Original NVidia mining code
 Authors: tsiv and KlausT
 License: GNU GPLv3
-Notes: Improvements are (c) of Xmr-Stak team team and are covered by GNU GPLv3
+Notes: Improvements are (c) of Xmr-Stak team and are covered by GNU GPLv3
 
 -------------------------------------------------------------------------
 
--- a/xmrstak/backend/globalStates.cpp
+++ b/xmrstak/backend/globalStates.cpp
@@ -35,57 +35,18 @@ namespace xmrstak
 
 void globalStates::consume_work( miner_work& threadWork, uint64_t& currentJobId)
 {
-	/* Only the executer thread which updates the job is ever setting iConsumeCnt
-	 * to 1000. In this case each consumer must wait until the job is fully updated.
-	 */
-	uint64_t numConsumer = 0;
-
-	/* Take care that we not consume a job if the job is updated.
-	 * If we leave the loop we have increased iConsumeCnt so that
-	 * the job will not be updated until we leave the method.
-	 */
-	do{
-		numConsumer = iConsumeCnt.load(std::memory_order_relaxed);
-		if(numConsumer < 1000)
-		{
-			// register that thread try consume job data
-			numConsumer = ++iConsumeCnt;
-			if(numConsumer >= 1000)
-			{
-				iConsumeCnt--;
-				// 11 is a arbitrary chosen prime number
-				std::this_thread::sleep_for(std::chrono::milliseconds(11));
-			}
-		}
-		else
-		{
-			// an other thread is preparing a new job, 11 is a arbitrary chosen prime number
-			std::this_thread::sleep_for(std::chrono::milliseconds(11));
-		}
-	}
-	while(numConsumer >= 1000);
+	jobLock.rdlock();
 
 	threadWork = oGlobalWork;
 	currentJobId = iGlobalJobNo.load(std::memory_order_relaxed);
 	
-	// signal that thread consumed work
-	iConsumeCnt--;
+	jobLock.unlock();
 }
 
 void globalStates::switch_work(miner_work& pWork, pool_data& dat)
 {
-	/* 1000 is used to notify that the the job will be updated as soon
-	 * as all consumer (which currently coping oGlobalWork has copied
-	 * all data)
-	 */
-	iConsumeCnt += 1000;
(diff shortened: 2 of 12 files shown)


fireice-uk:
This isn't correct. Consume needs to be per thread. Please explain in numbered steps how you think a deadlock can happen.

fireice-uk:
Also please note that a deadlock in worker threads wouldn't affect the network thread. I see some people already think it is some kind of client-side fix for nicehash having shitty software.

With regards to nicehash, I think we should silence them like cast-xmr but make it optional wit default to false.

THIRD-PARTY-LICENSES
```diff
@@ -5,7 +5,7 @@ This application bundles the following third-party software in accordance with t
 Package: Original NVidia mining code
 Authors: tsiv and KlausT
 License: GNU GPLv3
-Notes: Improvements are (c) of Xmr-Stak team
+Notes: Improvements are (c) of Xmr-Stak team team and are covered by GNU GPLv3
```
fireice-uk: > team team

THIRD-PARTY-LICENSES
```diff
@@ -27,3 +27,10 @@ Authors: okdshin
 License: MIT License
 
 -------------------------------------------------------------------------
+
+Package: cpputil
+Authors: Will Zhang
+Source: https://github.com/willzhang4a58/cpputil
```
fireice-uk: Nice to see. Big thanks for dealing with licensing and big thanks to @willzhang4a58 for the code.

xmrstak/cpputil/read_write_lock.h
```diff
@@ -0,0 +1,86 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef CPPUTIL_READ_WRITE_LOCK_H_
+#define CPPUTIL_READ_WRITE_LOCK_H_
```
fireice-uk: Can we get this to the standard stak format -> tab indent and `#pragma once`

xmrstak/backend/globalStates.cpp
```diff
@@ -33,24 +33,36 @@
...
-	// iConsumeCnt is a basic lock-like polling mechanism just in case we happen to push work
-	// faster than threads can consume them. This should never happen in real life.
-	// Pool cant physically send jobs faster than every 250ms or so due to net latency.
-
-	while (iConsumeCnt.load(std::memory_order_seq_cst) < iThreadCount)
-		std::this_thread::sleep_for(std::chrono::milliseconds(100));
+	jobLock.WriteLock();
+	
+	// this notifies all threads that the job has changed
+	iGlobalJobNo++;
```
fireice-uk: This is in a wrong place. The order that I told you:

when a global job is being written you:
acquire write lock

write the job
release the lock
increase the jobno

when a thread gets a job:
acquire read lock
copy the job
set the job no to global (!!! note the difference with write lock)
release read lock

Is slightly more efficient since the threads only start to read when the resource is available

psychocrypt:
updated

fireice-uk:
Needs a rebase before testing

psychocrypt:
I removed the BUG label. The point that all worker must consume a job before a new job can be added for the worker is no real bug. A drawback of this restriction can be that the miner is mining old jobs even if newer jobs are received but I do not know if this can be defined as BUG.

PR #1537, opened 2018-05-03
set correct `iPoolId` for miner_work

psychocrypt:
- initialize miner_work iPoolId in the default constructor with the invalid pool id
- move definition of `pool_data` into an own file (avoid ring includes)

If we lost the connection to the pool or at the start we set the miner_work to a default constructed work element. In this work the stalled flag is enabled and the pool id is set to default. In the previous code the id was default zero `0` which represent the dev pool. If we get the connection to a new pool or the old one after we lost the connection the previous pull in [executor.cpp:380](https://github.com/fireice-uk/xmr-stak/blob/26a5d65f12b2f19a0a3ece39a2bc64718796367b/xmrstak/misc/executor.cpp#L380) is used to store the last nonce. The nonce is wrongly passed to the dev pool.

I do not know if the old behavior is critical but it is not correct.

### DIFF
--- a/xmrstak/backend/globalStates.hpp
+++ b/xmrstak/backend/globalStates.hpp
@@ -1,25 +1,14 @@
 #pragma once
 
-#include "miner_work.hpp"
+#include "xmrstak/backend/miner_work.hpp"
 #include "xmrstak/misc/environment.hpp"
 #include "xmrstak/misc/console.hpp"
+#include "xmrstak/backend/pool_data.hpp"
 
 #include <atomic>
 
-constexpr static size_t invalid_pool_id = (-1);
-
 namespace xmrstak
 {
-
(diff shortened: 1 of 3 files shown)


fireice-uk:
Can you set it to invalid_pool_id instead of a magic number? https://github.com/fireice-uk/xmr-stak/blob/26a5d65f12b2f19a0a3ece39a2bc64718796367b/xmrstak/backend/globalStates.hpp#L9

psychocrypt:
Yes I can but than I need to move the devinition out of the current file
else we have a cycle include.

fireice-uk <notifications@github.com> schrieb am Fr., 4. Mai 2018, 11:35:

psychocrypt:
updated

PR #1543, opened 2018-05-05
Update webdesign.cpp - relative URLs

SoreGums:
When using a reverse proxy to expose stats to the internet.
Need to have relative URLs instead of absolute ones :)


SoreGums:
eh, didn't read :(

Issue #1527, opened 2018-05-01
Help needed for testing [only advanced users please]


psychocrypt:
If you* **know** how xmr-stak can be compiled it would be nice if you can test #1526 and report if you have any issues with this PR.
Please do not ask how you can compile xmr-stak, this issue is only for advanced users.

direct download: https://github.com/psychocrypt/xmr-stak/archive/fix-jobConsume.zip

The changes in #1526 should solve some possible race conditions during while the job is switch and possible deadlocks.

Spudz76:
Running this patch now on most of mine although I had no issues before (nanopool)
Should prove it doesn't cause any regressions.
Mix of cpus, amds, and nvidias, and mix of compilers/distros/drivers, too.

Amdpowered:
Compiled and running as well. No issue so far. But the error only happened once before this fix.

Spudz76:
This patch works identical to previous for me (meaning, no issues caused).  Installed on every rig now.

I get bad results occasionally (100/day) but believe it's from overclock or just old cards being old.  And quite a few expired/late results but that's a pool fixed-diff too high thing.  Still well under 1% and the exact same as earlier versions.

aij:
Just got back home to find 3 machines doing useless work again. Looks the same as before, but let me know if there's more I should be looking at.

![xmr-stak-b24cb3a2](https://user-images.githubusercontent.com/4732885/39553985-f9ab77b6-4e35-11e8-9402-46d2cde6fd81.png)

Spudz76:
I think that looks more like your Internet choked on itself and then the pool got permanently confused about your re-login for ~11 hours following.

aij:
@Spudz76 The tcpdump I [previously uploaded](https://github.com/fireice-uk/xmr-stak/issues/1505#issuecomment-385136163) made it pretty clear it's not the pool that got confused. How does this look different to you?

psychocrypt:
@aij  could you please download the branche https://github.com/psychocrypt/xmr-stak/archive/topic-consumeJobWithOutput.zip compile and run it.
This PR will create a lot of debug output. It would be nice if you can set a log file in config.txt and provide use with the log as soon as you reproduce the rejected issue.
Hopefully this will give us enough information to detect the source of the issue.

kaeptnb:
Hi! i tested this release because of issue #1519 - it helps cutting down the invalid shares from about 4-5% to 2% (but let me test this longer to be sure)

[...]

PR #1553, opened 2018-05-07
ignore duplicated equal job from pools

psychocrypt:
Some pools or proxys are sending multiple times in row the same job to the miner.
This PR will extent the miner to trigger an socket error if a job with the same `jobID` than the current active job is received.

Move motd evaluation before the possibility part where a job is ignored.

# Tests

- [x] disconnect after job is send twice

### DIFF
--- a/xmrstak/net/jpsock.cpp
+++ b/xmrstak/net/jpsock.cpp
@@ -399,18 +399,45 @@ bool jpsock::process_pool_job(const opq_json_val* params)
 		return set_socket_error("PARSE error: Job error 2");
 	}
 
+	if(motd != nullptr && motd->IsString() && (motd->GetStringLength() & 0x01) == 0)
+	{
+		std::unique_lock<std::mutex> lck(motd_mutex);
+		if(motd->GetStringLength() > 0)
+		{
+			pool_motd.resize(motd->GetStringLength()/2 + 1);
+			if(!hex2bin(motd->GetString(), motd->GetStringLength(), (unsigned char*)&pool_motd.front()))
+				pool_motd.clear();
+		}
+		else
+			pool_motd.clear();
+	}
+
 	if (jobid->GetStringLength() >= sizeof(pool_job::sJobID)) // Note >=
 		return set_socket_error("PARSE error: Job error 3");
 
-	uint32_t iWorkLen = blob->GetStringLength() / 2;
+	pool_job oPoolJob;
+
+	const uint32_t iWorkLen = blob->GetStringLength() / 2;
+	oPoolJob.iWorkLen = iWorkLen;
+	
 	if (iWorkLen > sizeof(pool_job::bWorkBlob))
 		return set_socket_error("PARSE error: Invalid job length. Are you sure you are mining the correct coin?");
 
-	pool_job oPoolJob;
 	if (!hex2bin(blob->GetString(), iWorkLen * 2, oPoolJob.bWorkBlob))
 		return set_socket_error("PARSE error: Job error 4");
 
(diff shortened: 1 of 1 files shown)


psychocrypt:
@fireice-uk Is it maybe better to add a method to `jpsock` to querry the jobid only to reduce the overhead of copying unused jobdata?

imperdin:
@psychocrypt It would be helpful to get a list of the buggy Pools / Proxies 
So each Pool / Proxy could fix these duplicate work issues locally

psychocrypt:
@imperdin a user saw it https://github.com/fireice-uk/xmr-stak/issues/1527#issuecomment-386914191 and it is not fully clear if it is the pool or proxy. This issue must be reported from the users. But an error message in the logs will help to report it.

xmrstak/net/jpsock.cpp
```diff
@@ -411,6 +424,17 @@ bool jpsock::process_pool_job(const opq_json_val* params)
 		return set_socket_error("PARSE error: Job error 4");
 
 	oPoolJob.iWorkLen = iWorkLen;
+
+	{
+		// lock reading of oCurrentJob
+		std::lock_guard<std::mutex> lock(job_mutex);
```
psychocrypt: @fireice-uk to you prefer `std::unique_lock`  over `std::lock_guard`?

xmrig:
* All **nodejs-pool** based pools sometimes sent duplicated jobs, simple ignore is enough.
* **dwarfpool.com** only reconnect is solution, also looks like this pool was ok sometime, but now broken again.
* **xmr.f2pool.com** need workaround this pool sent job in login response and right after it, again as regular job.

Actually check only `job_id` should be enough, at least in all known cases, but I compare full job.

xmrstak/net/jpsock.cpp
```diff
@@ -411,6 +424,17 @@ bool jpsock::process_pool_job(const opq_json_val* params)
...
+		// compare possible non equal length job id's
+		if(std::string(jobid->GetString()) == std::string(oCurrentJob.sJobID))
```
fireice-uk: So much copying for so little gain... `strcmp(jobid->GetString(), oCurrentJob.sJobID) == 0`

**Please** make sure sJobID is always going to end with `\0` otherwise both versions break.

psychocrypt: yes both are zero terminated. `oCurrentJob.sJobID` is max 63 characters and the full array (64 char) is zeroed. `GetString()` is giving a C string.

fireice-uk: This is not sufficient. This is after the compare. Please make sure that sJobID is also zeroed on initialisation.

psychocrypt:
I missed to add the  socket error, will do it now.

psychocrypt:
updated, added socket error

xmrstak/net/jpsock.cpp
```diff
@@ -399,20 +399,44 @@ bool jpsock::process_pool_job(const opq_json_val* params)
...
+
 	if (jobid->GetStringLength() >= sizeof(pool_job::sJobID)) // Note >=
 		return set_socket_error("PARSE error: Job error 3");
 
+	pool_job oPoolJob;
+
+	// lock reading of oCurrentJob
+	std::unique_lock<std::mutex> jobIdLock(job_mutex);
+	// compare possible non equal length job id's
+	if(strcmp(jobid->GetString(), oCurrentJob.sJobID) == 0)
```
fireice-uk: Please compare the blobs too. We only care about not sending duplicate nonce.

psychocrypt:
add blob compare

PR #1563, opened 2018-05-14
Updated .gitignore to ignore CMake files

spagno:
When you run cmake, it creates a lot of files which aren't in .gitignore

I added them

regards

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,9 @@ xmr-stak.kdev4
 cmake-build-release/
 cmake-build-debug/
 \.idea/
+
+# CMake files
+xmrstak/CMakeCache.txt
+xmrstak/CMakeFiles/
+xmrstak/Makefile
+xmrstak/cmake_install.cmake


psychocrypt:
never build directly in the folder xmrstak. Please always use a seperate
build folder or a tmp folder somewhere else.

psychocrypt:
I will not merge this PR because I need to fix this issue by disallowing the cmake call within the folder where the file CMakeLists.txt is located. I will keep this PR open until I fixed this issue

Spudz76:
There are a few errors in the build docs where it suggests `cmake .` (singular) not `..` and that may be misleading some cases.

[build-subdir-docs-patch.txt](https://github.com/fireice-uk/xmr-stak/files/2256543/build-subdir-docs-patch.txt)

Spudz76:
@psychocrypt also of note, this particular build was done in the source subdir (`xmr-stak/xmrstak`) not even in the base git checkout dir.  So you will need to check that if it is in-tree (not /tmp/ or such) then it is only 1 depth below the git-root but also not any of existing folders (`doc`. `CI`, `scripts`, `xmrstak`, probably even the `.github` or someone will build from there next somehow)

PR #1568, opened 2018-05-16
avoid out of order job processing

psychocrypt:
The login result of a pool contains the first job for the miner.
In the case where the pool is sending very fast after the pool login result
a new job it is possible that the newer job is processed faster than the job within the login result.
The result will be that the miner is mining an older job instead the newest (last received).

- enumerate all received messages
- trace the message id of the last procssed job
- skip all jobs where the message id is older than the last procesed id

# Test
- [ ] with mockup pool


psychocrypt:
updated: removed `this->`

PR #1535, opened 2018-05-03
Spell check

fireice-uk:

PR #1602, opened 2018-05-29
Update config.tpl

caimaoy:
modify sytems to systems


Spudz76:
Patch Requests are to be applied to the `dev` branch, like it says in multiple places.

Also, already fixed by #1535 a while ago, there just hasn't been a new release in a while.

psychocrypt:
thx for the spelling fix but like @Spudz76 mentioned it is already fixed in #1535

PR #1604, opened 2018-05-29
cpu cryptonight_twenty_hash

mbasaman:


mbasaman:
This seems to improve CPU performance a bit.  I'm surprised that the load takes so much time, I think the update makes the prefetch option more effective (at least on the CPU's that I tested).

mbasaman:
Yeah, that sounds good.

What was had observed is that 20 hashes at a time was faster than 3 or 5, but the optimal number might be 15 or something like that.

It's probably different for each machine.  I'm not sure exactly what the trade-offs are, but I think RAM access is the issue.

Spudz76:
`Intel(R) Xeon(R) CPU E5-2620 0 @ 2.00GHz`
15MB smartcache
Using Win7 with
```
    { "low_power_mode" : 20, "no_prefetch" : false, "affine_to_cpu" : 0 },
    { "low_power_mode" : 20, "no_prefetch" : false, "affine_to_cpu" : 1 },
    { "low_power_mode" : 20, "no_prefetch" : false, "affine_to_cpu" : 2 },
    { "low_power_mode" : 20, "no_prefetch" : false, "affine_to_cpu" : 3 },
```
and I assume that's correct to enable this patch (20 with prefetch not-disabled aka enabled)

Same config no patch with 1/true gave 278H/s and with patch is 261H/s a loss of 17H/s or 6.11%

-----
Atom D2550 with Linux and gcc-7 had a loss of half, but not surprising.  It isn't fast anyway and has no AES.  More testing out of curiosity, and it made it worse which is as expected.

-----
`Intel(R) Core(TM) i3-4160 CPU @ 3.60GHz`
As above Xeon test but with Linux and clang-3.8, otherwise same.
Normal was 86H/s and with patch is 81H/s loss of 5H/s or 5.81%

mbasaman:
I think the config has to be 6.

"cpu_threads_conf" :
[
    { "low_power_mode" : 6, "no_prefetch" : false, "affine_to_cpu" : 0 },
    { "low_power_mode" : 6, "no_prefetch" : false, "affine_to_cpu" : 2 },
],

It's faster on the machines I've tried

Spudz76:
That gave 60H/s on the little i3 under Linux, with or without the second host thread.  Waaay worse.
81H/s on the 278H/s Xeon E5 ewwww

mbasaman:
are you using low_power_mode == 6?  It's faster on several servers that I've tried

mbasaman:
One other thing to note is that I'm only using CPU 0 and 2, not 0, 1, 2 and 3.

Perhaps you can check if only using half of the CPU's has any effect.

mbasaman:
Yeah so, reducing the number of hashes to 10 and kicking it back up to 4 CPU's put it in line with 20 hashes on 2 CPU's.

[2018-06-01 06:03:39] : Starting (10)x thread, affinity: 0.
[2018-06-01 06:03:39] : hwloc: memory pinned
[2018-06-01 06:03:39] : Starting (10)x thread, affinity: 1.
[2018-06-01 06:03:39] : hwloc: memory pinned
[2018-06-01 06:03:39] : Starting (10)x thread, affinity: 2.
[2018-06-01 06:03:39] : hwloc: memory pinned
[2018-06-01 06:03:39] : Starting (10)x thread, affinity: 3.
[2018-06-01 06:03:39] : hwloc: memory pinned

Totals (ALL):     75.1   73.5    0.0 H/s
Highest:    79.0 H/s

Spudz76:
Yes the retest previous post was with == 6
I guess it's strange it accepts 20 if there is no index for such / must just default to 0/1 (same meaning) which explains the same result when using same `no_prefetch==true` as with 0/false/1
The == 6 result was clearly doing something different.
What compiler were you using, I am very likely using weird ones (clang, gcc-7) in comparison to yours.  Compiler may be optimizing out your hand expansions or such?

Spudz76:
I think I see, try to use less host threads (cores) so that the cache and prefetch stacks up longer queues (to hit that hundreds mark) instead of widening bandwidth and/or full cache size utilization (cache/2MB for monero7).  Taller work stacks via two blocks (4MB cache).

Had assumed your CPU was fourcore/octothread and for some reason your NUMA/hwloc placed cores interleaved with the HT (like Linux) thus the 0 2 4 6.  But you have octocore with 16-thread and are using half the physical cores so as to stack the prefetch higher.

Testing that...

mbasaman:
I'm not an expert on the compiler optimizations.  The compiler I'm using is: gcc (Debian 6.3.0-18+deb9u1)

If that doesn't work, you may want to try 10 hashes at a time of the full number of cores.  8 or 12 might be even better.

What I was thinking is that prefetch isn't helping as much as it should and the cause may be that there aren't enough CPU cycles after the prefetch command.

If you were reading from L1 exclusively I think it would be faster but there must be some trade-off with the limited number of registers.  Perhaps the store command is also a problem.

If it's not faster for you that don't worry too much about it.  Also, 10 hashes with the full number of cores may be more convenient to configure.

Spudz76:
I see around 65 per core with four host threads, 75 per core with two.
So there is a gain there, but using 4@65 is better than 2@75 and I can't seem to get 4@75.  BUT- this may be turbo kicking in because I am using less cores.
Otherwise various settings of low_power_mode even up to 3200 (didn't know it would go that high) gave more or less similar results (260H/s where normal was 268H/s) and then 151H/s with the two cores.  Would be real sweet to get the 302H out of these but again I think that is all due to the turbo from using 2 of 4 physicals. If I drop to one it should go slightly faster yet (80H/s?)

mbasaman:
What I would try is 10 hashes at a time.  You can change twenty_work_main() to use multiway_work_main<10u>(); instead of multiway_work_main<20u>(); and comment out the STEP(a10) through STEP(a19)

I'll make the change on the external branch or you can try it yourself.

mbasaman:
ok, I committed the changes to my local fork.  If you want to try it you can clone it.

It has similar numbers with 10 hashes on 4 cores to the 20 hash with 2 cores on this machine.

Spudz76:
Okay so what do you get with
```
"cpu_threads_conf" :
[
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 0 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 2 },
],
```

mbasaman:
I updated it to use 10 hashes (with the hope that you can use all 4 cores).

Try it a new build and use all 4 cores

"cpu_threads_conf" :
[
    { "low_power_mode" : 6, "no_prefetch" : false, "affine_to_cpu" : 0 },
    { "low_power_mode" : 6, "no_prefetch" : false, "affine_to_cpu" : 1 },
    { "low_power_mode" : 6, "no_prefetch" : false, "affine_to_cpu" : 2 },
    { "low_power_mode" : 6, "no_prefetch" : false, "affine_to_cpu" : 3 },
],

Totals (ALL):     76.3   74.7    0.0 H/s
Highest:    77.6 H/s

I think the base case of power level 5 was 68 H/s or something like that.

mbasaman:
Also, I just checked in a third update.

This to use the new code on power level 5 (as well as 6).  Perhaps the macro changes are the cause of the improvement I am seeing:

I had originally not wanted to touch the existing code.

"cpu_threads_conf" :
[
    { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 0 },
    { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 1 },
    { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 2 },
    { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 3 },
],

With the "latest" update to power level 5:

Totals (ALL):     76.3   75.9    0.0 H/s
Highest:    76.5 H/s

Before the update:

Totals (ALL):     63.6   67.7    0.0 H/s
Highest:    69.0 H/s

So perhaps the macro changes were the cause, not the increased number of hashes.  Check out the latest build and give *both* power level 5 and 6 a try with all 4 of the cores.

mbasaman:
Ok, so I made a fourth update to power levels 3, 4, 5, and 10 to use the branched code.

Starting 3x thread, affinity: 0.
Starting 3x thread, affinity: 1.
Starting 3x thread, affinity: 2.
Starting 3x thread, affinity: 3.

Original - Totals (ALL):     58.9   59.5    0.0 H/s
Updated  - Totals (ALL):     61.6   61.3    0.0 H/s

Starting 4x thread, affinity: 0.
Starting 4x thread, affinity: 1.
Starting 4x thread, affinity: 2.
Starting 4x thread, affinity: 3.

Original - Totals (ALL):     66.8   66.9    0.0 H/s    
Updated  - Totals (ALL):     72.6   72.4    0.0 H/s    

Starting 5x thread, affinity: 0.
Starting 5x thread, affinity: 1.
Starting 5x thread, affinity: 2.
Starting 5x thread, affinity: 3.

Original - Totals (ALL):     69.9   71.6    0.0 H/s
Updated  - Totals (ALL):     79.1   78.6    0.0 H/s   

Starting 10x thread, affinity: 0.
Starting 10x thread, affinity: 1.
Starting 10x thread, affinity: 2.
Starting 10x thread, affinity: 3.

Updated  - Totals (ALL):     79.8   78.0    0.0 H/s

Please update the build and give it a try.  I'm seeing it faster for power levels 3, 4, 5 and 10 now.  If that's not what you see, let me know.

*** note - I changed the config to expect 10 instead of 6

"cpu_threads_conf" :
[
    { "low_power_mode" : 10, "no_prefetch" : false, "affine_to_cpu" : 0 },
    { "low_power_mode" : 10, "no_prefetch" : false, "affine_to_cpu" : 1 },
    { "low_power_mode" : 10, "no_prefetch" : false, "affine_to_cpu" : 2 },
    { "low_power_mode" : 10, "no_prefetch" : false, "affine_to_cpu" : 3 },
],

Spudz76:
Okay, it looks like all the CPUs I was testing are Ivy Bridge variants, and none of this makes those faster at all, it hurts them bad.  Got 60/75 total versus the usual 260-280 across 3,4,5,10 leaving the affinity and thread counts the same as normal/best.

Tested this time on a Linux with weird [Intel(R) Xeon(R) CPU D-1518 @ 2.20GHz](https://ark.intel.com/products/91201/Intel-Xeon-Processor-D-1518-6M-Cache-2_20-GHz) which is Broadwell based and it did show gains though!
Normal best with false/true was ~92H and then with this patch and 10/false it gets 102H, running three cores of four, due to the 6MB cache (so only 3 x 2MB block space)

So maybe this just hates Ivy and Sandy memory controller design?  But I thought your CPU is an Ivy... did you try my CPU single thread no prefetch config on it from a couple posts back?  I'm curious if that roasts your other rates like it does on all of mine.

mbasaman:
This patch requires prefetch to be on.  Let me know what you'd like me to test.  Without prefetch it won't work.

Spudz76:
I want you to test my normal Ivy/Sandy config which I posted above, since you seem to have an Ivy core.  No low power and No prefetch, and yes then it's not using your code, but that is how I'm getting my rates that are 4x faster than with the prefetch and your patch.

Directed prefetch seems to work well with Broadwell cores, but however the Ivy/Sandy ones did their internal prefetch optimization (which can't be shut off) it doesn't like being directed, and I get 4x the performance by letting it do all prefetch management (aka "no prefetch" from xmr-stak / but again CPU still does its own regardless of that)

Spudz76:
Well I am wrong about the prefetches, according to [this](https://software.intel.com/en-us/articles/disclosure-of-hw-prefetcher-control-on-some-intel-processors) all those cores are "the same" as far as the optimizations they have and that they can be shut off.  I previously read somewhere that there are prefetches that are inherent and can't be shut off though... can't find it now.  But this Broadwell low power rig did gain with the patch which is what's confusing, why none of my others gain but instead lose by 4x.

I assume with no_prefetch:true this MSR is being set properly to turn them all off?  Maybe something with my motherboard or configuration is otherwise blocking it.

mbasaman:
Yeah, with a 4x loss I have no idea.

I'm not sure that I have that config.  The only config I see in your comments above is:

   { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 0 },
   { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 2 },

Totals (ALL):     27.3   24.7

The best config on this machine is:

    { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 0 },
    { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 1 },
    { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 2 },
    { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 3 },

Totals (ALL):     75.5   73.9

For the single thread tests:

Original

29.3 - { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 0 }
29.0 - { "low_power_mode" : 5, "no_prefetch" : true, "affine_to_cpu" : 0 }

26.7 - { "low_power_mode" : 4, "no_prefetch" : false, "affine_to_cpu" : 0 }
26.7 - { "low_power_mode" : 4, "no_prefetch" : true, "affine_to_cpu" : 0 }

Update

45.5 - { "low_power_mode" : 10, "no_prefetch" : false, "affine_to_cpu" : 0 }
21.6 - { "low_power_mode" : 10, "no_prefetch" : true, "affine_to_cpu" : 0 }

30.9 - { "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 0 }
24.1 - { "low_power_mode" : 5, "no_prefetch" : true, "affine_to_cpu" : 0 }

27.4 - { "low_power_mode" : 4, "no_prefetch" : false, "affine_to_cpu" : 0 }
24.2 - { "low_power_mode" : 4, "no_prefetch" : true, "affine_to_cpu" : 0 }

It appears that prefetch option had no effect in the original code (that's what I expected).

In the update, prefetch is better, but no prefetch is worse.  That loss can be refactored away.

However, if it is worse on your system, even with prefetch turned on, I think that would be a stopper.

Let me know if there is another config you want to test.

Spudz76:
Cool, strange.  How about that but on four cores, is it around 50H/s?
Therefore your choice of results being around 50 with no patch single thread no prefetch, around 64 with threading and prefetch no patch, and 75 with patch.  This does not jive with all my results which are 260, 64, and 75.

The Atom with 10 is hilarious and benchmarks at 650H/s.  But a real run on pool shows 0 and never seems to do anything but hear jobs and drool.

Spudz76:
Also I am essentially cherry-picking your patches and applying them on my dev-hax branch, so no need to revert yet, I can turn each part on and off as-needed by applying the patches in the other order or reverting some.

They do improve low_power_mode it's just that, on my rigs I see 4x even that performance just using one thread per 2MB cache.  And then why is that different between mine and yours if it's all Ivy Bridge and just varied cache size and core counts and HT or not.

Do you have the iGPU enabled and in use?  All of mine are either non-GPU (disabled from factory along with the HT on that weird Broadwell) or have the iGPU disabled (Supermicro with IPMI which brings its own dumb 2D video chipset).

Perhaps my magic performance is somehow using resources freed by killing iGPU and/or shared memory for it (also dis-mapped)?
It is weird because I get the same low_power_mode results as you do.  Thus I expect you should be hitting my same ~40H per configured core-thread. which would be like ~155H on the four solo-noprefetch setup.  But you get around 50 (projected from 2-thread @ ~25H)

Spudz76:
Got 42H -> 49H on this other `Intel(R) Core(TM) i3 CPU 530 @ 2.93GHz` with 4MB cache with patch and 10.  So some more gains, but that CPU sucks at this anyway (two threads because cachesize)

Got 166H -> 174H on a dual cpu `Intel(R) Xeon(R) CPU E5620 @ 2.40GHz` with 12MB cache with patch and 3, using 4 cores on each (8 total cores).  Was previously running 12 threads (to eat all cache) so two additional HT's on each cpu which was a couple hashes better than normal limiting at cores and not using 4MB of the cache... for some reason, it shouldn't be

Spudz76:
OK got to test this on Win7 `Intel(R) Core(TM) i3-4130 CPU @ 3.40GHz` with 3MB cache and found a strange way to new best of 106H/s total.  Use one thread with your code at 10, and another on the other physical core the way I usually have it set (false/true) and then both get around 53H/s each.

Not sure why/how (should be out of cache) but it works.  Normal with all false/true was 86H/s so it gained 20H/s.

Same gain same config in Linux on the i3-4160 3.6GHz 3MB.  Runs 6H/s slower overall in any mode so even with the faster CPU, Linux is slower.

```
    { "low_power_mode" : 10, "no_prefetch" : false, "affine_to_cpu" : 0 },
    { "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 2 },
```

mbasaman:
all right, well - I don't know if that is useful to you or not

If it's something that you want, what I would do is leave the existing methods exactly how they are ad add new methods for additional options.  I would add 6, 7, 8, 9 and 20 as options as well.  I think the only issue with this is adding to the compile time.

I have no idea why it works - as you said, it should be out of cache.  The only thing that I can think of is that will additional CPU cycles between the prefetch and the load, it is more likely that the memory is read from L1, which should be a lot faster than L3.

In addition to that, I would create a wrapper script that iterates through all the possible "combinations" to find the best config, by actually running for 1 or 2 minutes on each possible.

Spudz76:
Loaded that Xeon D-1518 Broadwell up with 10/false but used every core (0-7) which means HT cores also, and it scores 150H/s that way.  Up from original 92H/s, huge.

mbasaman:
are you comparing against the trunk?  I also changed 3/4/5

Spudz76:
no, against `dev` without your patches is the comparison (also the normal one I run daily)
I have never run `master` ever since I started compiling in-situ on each Linux rig.

And yes I have been checking all four permutations 3,4,5,10/false and have all the patches

mbasaman:
Ok, I'm running this script now

pid_t pid = fork();
if(pid == 0) {
	prctl(PR_SET_PDEATHSIG, SIGKILL);
	int stdout_fd = open("test.out", O_WRONLY|O_CREAT|O_TRUNC, 0644);
	dup2(stdout_fd, STDOUT_FILENO);
	close(stdout_fd);
	execl("xmr-stak", "xmr-stak", NULL);
	exit(1);
}
sleep(delay);
string hashrate = get_hashrate_from_log();
kill(pid, SIGKILL);
waitpid(pid, &result, 0);

...

test time: 75 seconds
threads: 4
power modes: 6
prefetch options: 2

testing 1819 configurations for 75 seconds
this should take about 37:53:45

saved cpu.txt to cpu.bak.2.txt
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
Hashrate: 15.9
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
Hashrate: 25.6
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
Hashrate: 28.3
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 3},
Hashrate: 29.4
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 2, "no_prefetch" : false, "affine_to_cpu" : 3},
Hashrate: 33.4
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 3, "no_prefetch" : false, "affine_to_cpu" : 3},
Hashrate: 35.3
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 4, "no_prefetch" : false, "affine_to_cpu" : 3},
Hashrate: 39.4
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 3},
Hashrate: 41.1
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 10, "no_prefetch" : false, "affine_to_cpu" : 3},
Hashrate: 47.0
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 1, "no_prefetch" : true, "affine_to_cpu" : 3},
Hashrate: 30.5
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 1, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 2, "no_prefetch" : true, "affine_to_cpu" : 3},

mbasaman:
It went from 72.4 to 86.1 on this i7-3537U CPU @ 2.00GHz

The script is checked in at xmr-stak-config.  You can take it and rename it if you want it.

power options: 1 2 3 4 5
branch: master

number of cores: 2
threads_per_core: 2
test time: 75 seconds
power options: 5
prefetch options: 2
configurations: 65
estimated runtime: 1:21:15

{ "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 5, "no_prefetch" : false, "affine_to_cpu" : 3},

Hashrate: 72.4

---

power options: 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 103 104 105 106 107 108 109 110 111 112 113 114 115
branch: more_options_dev

// N > 100 use the branched code optimized for prefetch with (N - 100) hashes

number of cores: 2
threads_per_core: 2
test time: 75 seconds
power options: 28
prefetch options: 1
configurations: 434
estimated runtime: 9:02:30

{ "low_power_mode" : 109, "no_prefetch" : false, "affine_to_cpu" : 0},
{ "low_power_mode" : 112, "no_prefetch" : false, "affine_to_cpu" : 1},
{ "low_power_mode" : 109, "no_prefetch" : false, "affine_to_cpu" : 2},
{ "low_power_mode" : 112, "no_prefetch" : false, "affine_to_cpu" : 3},

Hashrate: 86.1

mbasaman:
ok, I'm moving on to something else for the time being.  If you want to use the new 10 method, I would suggest keeping the unmodified 3/4/5 methods as was done in the 1st version of the checkin.

Good luck

Spudz76:
Thanks for all the work, I did not find the loop-test-script on your fork anywhere, but will probably write my own to use `--benchmark 7 --benchwait 1 --benchwork 14` so it only wastes ~15s each test, and doesn't need to connect to a pool at all.

I would have been unable to use the test script as my Internet doesn't always connect to pool on the very first packet / hangs and requires a kick / no way to detect that easily in looper script.  Also benchmark mode exits by itself (no sleep/kill).

mbasaman:
yeah, it's no problem at all.  I'll probably read the GPU script as well eventually, it's probably already good, but you it's always possible that there is something there.

The script is at https://github.com/mbasaman/xmr-stak-config

You can update the code to pull the 15 second value.  It's just scraping the log file.

I think what it's doing now is waiting 75 seconds, then if the log entry doesn't get generated, it will re-read the log every 5 seconds for an additional 75 seconds.  You can update it to wait longer for the network issues.

psychocrypt:
Could someone please give me a short summery if 20hash is an improvement and short summarize an example. Please do not measure one core. Please use a good old cfg vs a cfg with this PR.

mbasaman:
if you want to use the "10 hashes at a time" method OR the "optimized for prefetch" changes, I would suggest a new PR that retains the original implementations.

https://github.com/mbasaman/xmr-stak/tree/more_options_dev is an example, but it would need to be trimmed to improve build time.

The last tests I did are available at:

https://github.com/mbasaman/xmr-stak-config/blob/master/results.2.txt

which showed a 20% improvement versus

https://github.com/mbasaman/xmr-stak-config/blob/master/results.1.txt

psychocrypt:
what kind of cpu do you use for your tests? Leds than 100H/s looks like
some very old low end cpu.

Spudz76:
We were pretty clear everywhere above about which CPUs got what and we tested quite a range

And yes like Celerons and stuff that aren't supposed to work well anyway.  But the main interesting thing is how stacking silly high levels of work lets the Intel funky magic (SmartCache, predictive whatever, adjacent cache line prefetch, etc...) do a much cleaner job which gets quite a boost in some cases.

I think once a huge stack of tasks is queued it stops readjusting the SmartCache topology as much (as it can see a gigantic workload of the same stuff over and over) but when it remains a normal logical task with short/fast sections, it decides to share out the cache to cores differently, and changes it often - leading to not same hashrate on each core (cores should be very close to completely equal) or floating hashrates ('high' rate core floats among the others, but overall total hashrate is stable).

Since SmartCache is relatively snake oil with no docs, it makes sense to try semirandom ideas just to see how it reacts, which might tell us more about exactly how to force (or pseudo-force) cache-to-core allocations.

Spudz76:
#1649 may be related to SmartCache being "weird" with too-short a work stack, but definitely shows the problem, and I guess as usual "some other miner" is perfect every time.

Spudz76:
It probably needs a thorough cleanup, and/or it seems like a template could generate a variable selection of `low_power_mode` thread depths.  When compiling with 2-150 it takes about an hour, building a kernel per thread depth, and the executable is gigantic, so including them all is not particularly great except for benchmarking to figure out which depths to include for your system(s).

Would be better to utilize hardware ID information and have a DLL per family, split the CPU backend into so/dll like the other backends.  Memory footprint would be smaller, loading at runtime whichever CPU backend has whatever heights work best on it.

But collecting reliable benchmark data is somewhat a problem, so a profiler tool would have to be created (`xmr-chek` ?) to use most of the core backend code but only run hardware ID and benchmarks (test all possible backends and step through thread depths for 5s each and then retest the top ones for 15s each like successive search) and provide a file to post/upload somewhere, in a machine importable format.  That would provide enough information to know what's best for a particular CPU and/or family of CPUs, which then could be used to make a specific backend for that ID/fingerprint (model/stepping/revision/cache amount/etc).

Bench-cloud system could benefit the autoconfig code in the GPU backends as well, not having a benchmark-cloud type system makes it tough to get good optimization data for those just the same.

There is that site with various paste dumps of old format config files and whatnot, but that isn't machine readable nor always accurate and most are from ancient backends or old algorithm versions (out of date).

mbasaman:
yeah, so it does need to be cleaned up.

There are probably ways to reduce compile time by moving the code to a static library (or something similar).

What I was thinking is to add some kind of learning process to the code if every CPU is different.  For example, you mine for 9 minutes on the best known config, and then spend 1 minute mining on an untested config.  After a few hours or days, the program will have built it's own deployment-specific benchmarks, which could be persistent locally via the file system or whatever.

After the benchmarks have been auto-generated (after a few days) it could still try to re-benchmark known configs but probabilistically favor configs that are almost as good as the best config, on the off chance that there was some irregularities during the original bench-marking.

That's assuming that you have lots of combinations of implementations via static libraries.  The other option would be to just add a single method with 10 hashes per run using the modified macros that are optimized for prefetch and leave the original code unmodified.

mbasaman:
and if you guys don't have time to put it together yourself, just agree to a specification for what you want done and I'll implement it when I have some free time

baldpope:
I like @Spudz76 suggestion above of running a benchmark and uploading results some how.  I agree, the current xmr bench mark site is OK as a reference, but far from definitive and not machine readable at all.

I'm not familiar with the client side portion at all, but could put take a stab at putting something on the get/share results side.  What all field data would be relevant to store?

Spudz76:
It would be ideal if the CPU backend compiled its executable/kernel at runtime based on settings/detections just like the GPU backends.  But then the miner rig needs compilers, not a problem for Linux but, windows would be painful.  However it would solve the huge-exe or many-dll problem and allow for one piece of code with optional sections and variable template expansions/unrolls just like the OpenCL or CUDA code does, assembled at runtime.

Maybe this is why OpenCL supports CPUs at all, sort of a portable compile engine.  Too bad it is never fast as direct to metal, but I do wonder how close it can get (one would hope Intel or AMD made their CPU-OpenCL with strengths and weaknesses of their CPU in mind)

Spudz76:
I've noticed my current branches with the 10-way patch applied probably do not work correctly for anything but what I mine (monero7) there are some missing patches to the macros for newer/changed coins.

Caveat until I get them fixed up.  Maybe converting the macros to true templates too.

I did modify and succeed in getting the cpu backend compiled into a shared library just like the other two.  There is some mess with the other backends needing to use the 'core' cpu code to validate results, but I forked a simplified singlethread CPU hash validator to leave in the main backend static lib for those to all use.  Bonus there is the unoptimized validator should never have any optimization-related miscalc problems, and not contend with CPU mining that is using cache (the validator is run only when one of the backends reports a result, not continually in a loop, so speed is not an issue and it can stay out of the way).

I am currently wrestling with CMake not building the OpenCL backend until after the main exe, so then it doesn't get linked properly.  CUDA still builds before the main exe so I've got something messed up.

But CPU mining via the externalized backend seems to work so far.

Spudz76:
Also suspicious that the X-way cache on the CPU may be the precise X-way threading that works optimally.  One of my CPUs has 20-way 15MB L3 thus the hypothesis is that 20-way threading will work best.

Pretty much rewriting the autotuner to use more knowledge that hwloc provides, and finally merging the hwloc and non-hwloc code into one file so it's easier to work with.  The non-hwloc autotuner also had more messages about what it found and how it thinks the layout should be and why, which is useful info but totally missing in the hwloc version.

mbasaman:
ok, I'm glad you picked it up. good luck

PR #1585, opened 2018-05-20
Warning on Intel C/C++

jdmedeiros:
Avoid a warning if using Intel C/C++

Please make sure your PR is against **dev** branch. Merging PRs directly into master branch would interfere with our workflow.

PR #1607, opened 2018-05-30
remove definition of `_mm256_set_m128i`

psychocrypt:
This is a replacement PR for #1585, thank you @jdmedeiros for pointing us to this code snipped and providing a fix for the intel compiler.

The define of `_mm256_set_m128i` is not needed and avoid intel compiler throws a warning about it.


jdmedeiros:
Thanks @psychocrypt for picking up the ball.

PR #1609, opened 2018-05-30
fix broken AMD APP SDK download links

psychocrypt:
AMD has removed the AMD APP SDK links from the website. The AMD employes in the AMD formum posted links to there Dropbox where the SDK can be downloaded)


dormancygrace:
Their SDK now lives there https://github.com/GPUOpen-LibrariesAndSDKs/OCL-SDK

PR #1637, opened 2018-06-05
Add Cryptonight-fast - Masari PoW variant

gnock:
Masari is forking to a variant of cryptonight variant v1 with less iterations, called cn-fast.

The implementation has been tested on testnet.masaricoin.com with cpu/nvidia/amd miners.

The fork date is set to be around Monday June 18th.


psychocrypt:
Is the coin forking on bloch version 5 or 6? Because the coin is already
forked from cn to cn-v7 with block version 5 but thr code shows it is again
using the version 5.

psychocrypt:
please provide a testnet wallet address that i can test it too.

psychocrypt <psychocrypthpc@gmail.com> schrieb am Di., 5. Juni 2018, 18:10:

> Is the coin forking on bloch version 5 or 6? Because the coin is already
> forked from cn to cn-v7 with block version 5 but thr code shows it is again
> using the version 5.
>

gnock:
You can use that address as a testnet address : 6b52MV67y9y8jpaqPiB25u6CtGf7mUuzgf95roVhhxnkC9mLQC3NmSDjEswPNWjkCPBcDohFbc2MKLLC96E15cSpGqWicE8

in the current config, it works on the testnet pool & other pools (non-testnet), but we are checking the variant and making tests as it may be wrong. Thx for the catch !

psychocrypt:
But I can not accept the PR as long as the block version to fork is not
changed. The reason is that the miner can only detect the fork if the
version is changed.

gnock:
Ho i may not have been clear enough in my text sorry. I made the modification in the text and forced a push to keep it in 1 commit. It's now on 7u as it should be.

psychocrypt:
I renamed `cryptonight_fast` into `cryptonight_masari`, I wait now for the CI to pass

psychocrypt:
forget to note: the hash rate will go a little bit down each time the dev pool is mining. The reason is that the dev pool will stay mining cryptonight_monero. Do not woory about that this will have no impact on the user pool hashrate

psychocrypt:
binaries are available in the newest relese https://github.com/fireice-uk/xmr-stak/releases/tag/2.4.5

netxecute:
when you run stak 2.4.5 first time, it says use cryptonight_masari, but I think that forces the new algo.
if I use just masari will it switch automatically? (it doesn't mention it specifically like it does for haven in the pools.txt)

PR #1648, opened 2018-06-08
Re-read pool config while miner is running

syncro2017:
Wanted to have the option to edit and reload the pool config file without stoping the miner.
Tested and it seems to work, all your input and work is greatly appreciated.


syncro2017:
I just created the new pull request from the dev branch as you have instructed. thank you

PR #1639, opened 2018-06-06
Add support for CryptoNight Haven (small Heavy tweak)

havenprotocol:
Haven is hardforking in the next few weeks and making a small change to the algo in order to remove nicehash. The patch contains that tweak and allows the miner to autoswitch once we hit V3.

PR #1652, opened 2018-06-10
Add support for CryptoNight Haven (small Heavy tweak)

psychocrypt:
This PR replace #1639

- update pools.txt
- add new algorithm `cryptonight_haven`
- update all backends

# Tests

- [x] CPU (low_power false, 2,3,4,5)
- [x] AMD
- [x] NVIDIA


newmzy:
In your notes it says "Add support for CryptoNight Haven (small Heavy tweak)" - does this mean you have 'tweaked' the CryptoNight_Heavy algo generally or only when it applies to Haven mining?

Just wondering because I mine Loki - also uses Heavy & I have stability issues using different Nvidia 10xx GPUs on the same rigs.

jaredkaragen:
[https://github.com/fireice-uk/xmr-stak/pull/1652/files/b55acb719b933a5df9812f7af440cd3fc5c57da2](url)

@newmzy you can clearly see in the source changes in the link above it has conditionals for both algos....   Its safe to assume that it is separately done as can be seen by the different code.

psychocrypt:
Haven derived there POW function from cryptonight heavy. You can still
select heavy but also haven

PR #1655, opened 2018-06-10
Add support for reloading the pool config file while miner is running

syncro2017:
re-based changes against the dev branch as was instructed


xmrstak/misc/executor.cpp
```diff
@@ -600,7 +622,7 @@ void executor::ex_main()
 	if(jconf::inst()->GetVerboseLevel() >= 4)
 		push_timed_event(ex_event(EV_HASHRATE_LOOP), jconf::inst()->GetAutohashTime());
 
-	size_t cnt = 0;
+	size_t cnt = 0,i=0;
```
psychocrypt: please remove `i=0` it looks like it is not needed

xmrstak/misc/executor.cpp
```diff
@@ -551,42 +535,80 @@ void executor::ex_main()
...
+	
 	switch(jconf::inst()->GetCurrentCoinSelection().GetDescription(0).GetMiningAlgo())
 	{
-	case cryptonight_heavy:
-		if(dev_tls)
-			pools.emplace_front(0, "donate.xmr-stak.net:8888", "", "", "", 0.0, true, true, "", true);
-		else
-			pools.emplace_front(0, "donate.xmr-stak.net:5555", "", "", "", 0.0, true, false, "", true);
-		break;
+		case cryptonight_heavy:
```
psychocrypt: could you please remove all style changes that we can focus on the code.

xmrstak/misc/executor.cpp
```diff
@@ -487,26 +488,8 @@ void disable_sigpipe()
...
-	if(pvThreads->size()==0)
-	{
-		printer::inst()->print_msg(L1, "ERROR: No miner backend enabled.");
-		win_exit();
-	}
-
-	telem = new xmrstak::telemetry(pvThreads->size());
-
-	set_timestamp();
+void executor::init_pools(void)
```
psychocrypt: there is not need for the explicit `void`

psychocrypt:
@fireice-uk please have a look to it, do you have any arguments against this functionallity?

xmrstak/misc/executor.cpp
syncro2017: I believe this is fixed now

xmrstak/misc/executor.cpp
syncro2017: I have removed explicit void, however coming from embedded development for mission critical apps i always put void in the func that is not supposed to take any args, but this is in C :)

xmrstak/misc/executor.cpp
syncro2017: i is needed here because this variable is being used later on in one of the case statements in 2 places:

case EV_PERF_TICK:
                        for (i = 0; i < pvThreads->size(); i++)
                                telem->push_perf_value(i, pvThreads->at(i)->iHashCount.load(std::memory_order_relaxed),
                                pvThreads->at(i)->iTimestamp.load(std::memory_order_relaxed));

                        if((cnt++ & 0xF) == 0) //Every 16 ticks
                        {
                                double fHps = 0.0;
                                double fTelem;
                                bool normal = true;
                                for (i = 0; i < pvThreads->size(); i++)
                                {
                                        fTelem = telem->calc_telemetry_data(10000, i);
                                        if(std::isnormal(fTelem))
                                        {
                                                fHps += fTelem;
                                        }
                                        else
                                        {
                                                normal = false;
                                                break;
                                        }
                                }

                                if(normal && fHighestHps < fHps)
                                        fHighestHps = fHps;
                        }
                        break;

psychocrypt:
please define the variable 'i' in the for loop. defining all variables at
the beginning is old C style.

syncro2017:
Moved variable i counter definition inside the loops as requested

PR #1660, opened 2018-06-14
Grammar fixes; mmap alloc failed error remark for Arch Linux.

murlakatamenka:
Just a few grammar fixes in the docs.
Also a remark on fixing `mmap alloc failed` on Arch Linux and its derivatives. It should also be relevant for [Fedora](https://ask.fedoraproject.org/en/question/41559/has-etcsysctlconf-been-deprecated-on-fedora-20/) but I'm not its user (and didn't bother checking via VM or docker), needs confirmation from actual ones.

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -35,7 +35,7 @@ Reference: http://rybkaforum.net/cgi-bin/rybkaforum/topic_show.pl?pid=259791#pid
 
 If you set up the user rights properly ([see above](https://github.com/fireice-uk/xmr-stak/blob/master/doc/FAQ.md#selockmemoryprivilege-failed)), and your system has 4-8GB of RAM (50%+ use), there is a significant chance that there simply won't be a large enough chunk of contiguous memory because Windows is fairly bad at mitigating memory fragmentation.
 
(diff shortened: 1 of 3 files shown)


Spudz76:
The Arch instruction is generally correct for all distros, main monolithic sysctl.conf has been deprecated for a while more or less globally, or at least the newer subfolder method had been enabled so it works either way even on old distros.

Also less clashy with package manager updating config files which was the whole reason for the subfolder and locally named files that won't exist in the distro package.

The limits stuff also has a `.d` subfolder and is technically outdated methodology in the docs as well.

I personally also install sysfsutils and use its `.d` folder to setup power management and process scheduling for CPUs which isn't even in the docs (it doesn't make a huge difference).

murlakatamenka:
@Spudz76 thanks for clearing this out! I wasn't sure so decided to put a remark in a safe manner.
Yes, with a folder method it's definitely easier to deal with config files.

I'll update the doc a little bit taking into account your comment.

psychocrypt:
I will merge the PR as it is if there are further changes needed please open a new PR.

PR #1672, opened 2018-06-21
Buffer output in each line

aledomu:
Right now the output is buffered with the libc default policy if `flush_stdout` is set to false. This means that if the output is the terminal (`stdout`) it will be line-buffered (except in Windows, where it's unbuffered), but if it isn't then it will be block-buffered, which becomes too slow to pipe to other program or a file. Note that when `flush_stdout` is set to true, stdout is flushed **twice** on each line.

Since those are all the possible use cases, with this PR the output will be line-buffered and flushed **only once** in all cases and therefore no config entry will be necessary.

Edit: The behaviour of Windows caught me off-guard, so I had to revert the initial modification I made, force full-buffering for everything and do the flush manually. Now it works the same across all OS. I will keep my opinions on this to myself.

### DIFF
--- a/xmrstak/config.tpl
+++ b/xmrstak/config.tpl
@@ -114,14 +114,6 @@ R"===(
 "daemon_mode" : false,
 
 /*
- * Buffered output control.
- * When running the miner through a pipe, standard output is buffered. This means that the pipe won't read
- * each output line immediately. This can cause delays when running in background.
- * Set this option to true to flush stdout after each line, so it can be read immediately.
- */
-"flush_stdout" : false,
-
-/*
  * Output file
  *
  * output_file  - This option will log all output to a file.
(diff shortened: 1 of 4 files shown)


psychocrypt:
I assigned @fireice-uk because he wrote the output stuff.

fireice-uk:
The point of this option is to flush after each logical print. What would be the benefit of removing it?

aledomu:
Make it the default behaviour in every case, because not doing it doesn't make sense in any situation to my knowledge.

fireice-uk:
We added that option to work around specific problems with pipes - how do you address them?

fireice-uk:
@aledomu Thanks for the explanation! One more request, can you squash your PR? I don't like to have dozens of commits per pr unless the issue warrants it.

aledomu:
Done. I squashed the commits I should have but forgot about it, sorry. I prefer not to bundle everything in a single commit for better traceability and future understanding of modifications, and I think 3 commits are manageable. If you still want to do so, don't hesitate to ask.

PR #1703, opened 2018-07-08
add ryo currency

psychocrypt:
- add ryo currency
- update documentation


fireice-uk:
@psychocrypt Can we remove sumokoin as they are no longer GPU mineable?

PR #1704, opened 2018-07-08
remove cryptonight coins

psychocrypt:
- remove all cryptonight coins (algorithm cryptonight can still be used)
- update README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,7 @@
 ###### fireice-uk's and psychocrypt's
-# XMR-Stak - Monero/Aeon All-in-One Mining Software
+# XMR-Stak - Cryptonight All-in-One Mining Software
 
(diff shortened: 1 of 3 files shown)


fireice-uk:
I would opt to remove them altogether - we have pretty much zero cn-original donations, what's the point of maintaining unused coins?

Spudz76:
If more coins besides Electroneum switch back to the cn-original version due to [CG]PU miners doing too much profit switching and leaving the network slow on transaction processing, then it is still potentially valuable in the future, not necessarily deprecated.

psychocrypt:
I removed all native cryptonight coins

fireice-uk:
@Spudz76 We are not removing the algo, that remains, we are removing the named coin support for those coins.

PR #1732, opened 2018-07-20
Add OCL libraries to the project

dbielecki82:
Hello,

As title mentioned those libraries are quite light and apparently not being updated at all from quite some time.

I've found other official place which you can get OCL libs for the project buth install path is slightly different thus I though It wouldn't be a bad idea of including those libraries already into the project.

Libs were taken from here:
https://github.com/GPUOpen-LibrariesAndSDKs/OCL-SDK/releases 

On the right side of the screen you can see default installation path, on the left xmr-stak-dep with those libs copied into CL folder:
http://prntscr.com/k91g4w

This is what has to be added into CMAKE_PREFIX_PATH so finally it will look like this CMAKE_PREFIX_PATH = C:/xmr-stak-dep/hwloc;C:/xmr-stak-dep/libmicrohttpd;C:/xmr-stak-dep/openssl;c:/xmr-stak-dep/CL

Folder includes this:
[cl.zip](https://github.com/fireice-uk/xmr-stak/files/2215464/cl.zip)


Spudz76:
I was originally going to complain but

AMD actually says to use this same garbage repo, so as unofficial as it seems I guess it's the best headers now?

Have not tested a compile using them yet.

Not sure what `opencl-headers` or `opencl-amdgpu-pro-dev` are for then... as they also provide headers one from Ubuntu official and one from the amdgpu-pro 18.xx drivers.

In all cases the library comes with the driver.

psychocrypt:
sry this is not an pull request this is an request which must be an issue. Never the less the opencl libs provided via the miner is not an good idea because than we need provide windows, linux and MacOSX libraries. Also you should always use th elibs provided from you driver.

Spudz76:
@psychocrypt actually this is important, AMD points to these (sometimes) as the official headers to compile against

there are no libs/binaries those only show up with the drivers -- they pack the source in a zip for more or less confusing reasons / not really using github correctly
this is just the "new" Khronos official set of OpenCL SDK headers and there is no way to install them in windows, so similar to the rapidjson in-tree clone we should have these headers and compile against them unless overridden.  Pretty much adapt/import [this commit](https://github.com/xmrig/xmrig-amd/commit/ff04486d9e8859be3b6707ccd3405970c2c98460) because that looks super legit and is why the new drivers seem to work good with that.  Also makes the so/dll load happen fully at runtime so zero linking even... which seems super compatible?  It loads the library and then maps all the function entries into a vector table, pretty slick, and completely drops the linking requirement to the library at compile time.  May also fix incompatibility in Windows with Intel iGPU OpenCL when installed with AMD OpenCL (usually crashes/segfault currently, but seems to depend on which ICD is first platform) and maybe also the Linux Clover (when installed) crashing for nearly the same reason I suspect.

I have to test but these should also work backward, in that the APP-SDK is not needed even with 17.x or whatever else versions of driver just these headers.  The [docs say](https://github.com/xmrig/xmrig-amd/commit/ff04486d9e8859be3b6707ccd3405970c2c98460#diff-66d5217775fe2ae4df43efcbdbe9ea1bR14) "unified" versions in this pack, so now you define what version prior to the include of the main header, and then it supports that version (so the define would be based on the OpenCL library version from the driver, to match the header "personality" needed to work with it) and apparently this works for every OpenCL in the universe...

And that commit is this same pack of headers from that repo linked above, which is official even though it looks half hearted and fake and they seem to be using github wrong (just checkin the src guys?)

PR #1747, opened 2018-07-25
 First attempt at MoX support

alfredmox:
Hello,
We launched MoX few days ago with a new variant of CryptoNightV7 which consists of CNLite and a tweak variant.
Here is my first attempt (not valid yet) at integrating it into xmrig

Our github : https://github.com/mox-project/MoX
and here is the tweak in question : 
```
#define VARIANT1_1(p) \
  do if (variant > 0) \
  { \
    const uint8_t tmp = ((const uint8_t*)(p))[11]; \
    static const uint32_t table = 0x75310; \
    const uint8_t index = (((tmp >> 8) & 6) | (tmp & 1)) << 1; \
    ((uint8_t*)(p))[11] = tmp ^ ((table >> index) & 0x30); \
} while(0)
```
I think the bitshift I used in the opencl / cuda part is wrong as well

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/gpu.cpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.cpp
@@ -1004,7 +1004,7 @@ size_t XMRSetJob(GpuContext* ctx, uint8_t* input, size_t input_len, uint64_t tar
 		return(ERR_OCL_API);
 	}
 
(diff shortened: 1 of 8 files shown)


psychocrypt:
Currently I can not find any information about MoX. If I visit the github repo and follow the link https://getmox.org/ I end up with an error. If it is a cryptonight_lite derivative the coin should have at least something like 10MH/s before we will include it in the miner.

alfredmox:
Ok thanks for your answer, I'll come back to that pull request once we have a bigger network

psychocrypt:
Thanks for the PR as I wrote the projects needs at least 10MH/s and a working website. Please open a new one if the requirements are fulfilled.

PR #1752, opened 2018-07-27
Optimized AES Instructions for AMD

imperdin:


Spudz76:
Test results of any gains?

Pretty sure the compiler optimizes down to that exactly the same anyway, so besides making it less readable there is no difference.  Similar to hand-unrolling sections the compiler will unroll for us, it feels like an improvement but really nothing happened.

If it does indeed produce better machine code, then I would enjoy being wrong.

psychocrypt:
IN the past the code was exactly what you now provide. I changed it because I saw a small increase in the hashrate if I change it to what it currently is.
Could you please post the output of `xmr-stak --benchmark 7 --currency monero7 --noCPU --noNVIDIA` with this PR and with the current dev.
@Spudz76 could you please also test it on your VEGA. I have no VEGA system and if I remember @imperdin said me that there is only a increase for VEGA GPUs.

imperdin:
It's a little boost nothing that is really massive in terms of H/s gains.  

I've and received about a ~0.5% boost in Vegas
About ~3-5 H/s increase for Polaris 

While it's not the most dramatic boost, a boost is a boost
And it's more readable as well

My fork of Stak have that change and it seem to be stable and working on all AMD cards

Spudz76:
Seems like no difference to Tahiti on `17.50-552542` and no difference to Redwood or Tahiti on `fglrx 13.152`

Hawaii, Fiji, and Ellesmere are on Ethereum now but next time I bring them down for maint I'll check those too (same 17.50 as above so I doubt there will be any difference)

This [post on ROCm github](https://github.com/RadeonOpenCompute/ROCm/issues/132#issuecomment-312926987) contains what I believe the difference to be: if you run the legacy OpenCL then this patch does nothing and compiles to identical machine code.  If you run ROCm then it has that new style compiler that works differently, and this patch seems to have different outcomes then.  Probably why nearly everything is slower on ROCm, that compiler needs lots of help to be as good as the old Pro one, and it isn't up to speed yet.

Spudz76:
I think after reviewing [this fork](https://github.com/xmrig/xmrig-amd/blob/dev/src/amd/opencl/wolf-aes.cl#L90) I like the way it is done there, local `uint4` with the batch xor at the end.

imperdin:
In my unreleased Alpha version this is how I implemented my version of Short rounds

PR #1766, opened 2018-08-04
Update compile_Linux.md

timomwa:
Added -DCUDA_ENABLE=OFF -DOpenCL_ENABLE=OFF to cmake


psychocrypt:
your pull request is only valid for cpu. The way how it is desciped in the
current docs is correct. If you have no OpenCL or CUDA than cmake will
guide you what you must do to compile the miner.

psychocrypt:
closed because this would avoid GPU usage.

Issue #1780, opened 2018-08-13
bug: xmr-stak  masari mining broken （result rejected）


hieixu:
Please provide as much as possible information to reproduce the issue.

low_power_mode set to 2 cause the bug.
fine when low_power_mode set to false

# Basic information
CPU only minning

# Compile issues
  - Which OS do you use?
  ```
Debian 9 X64
  add **all** commands you used and the **full** compile output here
  ```
  ```
./xmr-stak
  run `cmake -LA .` in the build folder and add the output here
  ```
-- The C compiler identification is GNU 6.3.0
-- The CXX compiler identification is GNU 6.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CUDA_TOOLKIT_ROOT_DIR not found or specified
-- Could NOT find CUDA (missing:  CUDA_TOOLKIT_ROOT_DIR CUDA_NVCC_EXECUTABLE CUDA_INCLUDE_DIRS CUDA_CUDART_LIBRARY) (Required is at least version "7.5")
CMake Error at CMakeLists.txt:199 (message):
  CUDA NOT found: use `-DCUDA_ENABLE=OFF` to build without NVIDIA GPU support

-- Configuring incomplete, errors occurred!
See also "/home/xmr/src/xmr-stak-2.4.6/CMakeFiles/CMakeOutput.log".
-- Cache values
CMAKE_AR:FILEPATH=/usr/bin/ar
CMAKE_BUILD_TYPE:STRING=Release
CMAKE_COLOR_MAKEFILE:BOOL=ON
CMAKE_CXX_COMPILER:FILEPATH=/usr/bin/c++
CMAKE_CXX_FLAGS:STRING=
CMAKE_CXX_FLAGS_DEBUG:STRING=-g
CMAKE_CXX_FLAGS_MINSIZEREL:STRING=-Os -DNDEBUG
CMAKE_CXX_FLAGS_RELEASE:STRING=-O3 -DNDEBUG
[...]

PR #1821, opened 2018-09-10
unify cpu cryptonight implementations

psychocrypt:
xmr-stak has several implementations for multi hash per thread.
The results into 3 intepedent implementations.
Each time the algorithm must be changed the possibility to introduce errors is very large.

- unify the different cryptonight CPU implementations
- simplify the function selection array to find the specilized cryptonight implementation

Note: I tried to implement all as recursive unrolling with templates because I am no fan of macros. The result was a little bit slower code. Therefore I decide to use the ugly macros to implement it.
The new code is on my laptop `Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz /4MiB cache)` for cryptonight_lite with 4 hashes per thread slidly faster than the old implementation.
Never the less I will test a few more CPUs.

The reason why I have not reused the old macros is that the variables was defined outside of the macros. This is again a point where errors can be introduced very fast e.g. during the implementation of cryptonight_v8. After this PR we get always correct results if the single hash implementation is correct.

This PR solves also #1780

- [x] Test on different CPUs  
- [x] windows tests
- [x] test NVIDIA and AMD (to avoid that the CPU check of the GPU nonce is broken)

### DIFF
--- a/xmrstak/backend/amd/minethd.cpp
+++ b/xmrstak/backend/amd/minethd.cpp
@@ -252,7 +252,7 @@ void minethd::work_main()
 
 				*(uint32_t*)(bWorkBlob + 39) = results[i];
 
-				hash_fun(bWorkBlob, oWork.iWorkSize, bResult, cpu_ctx);
+				hash_fun(bWorkBlob, oWork.iWorkSize, bResult, &cpu_ctx);
 				if ( (*((uint64_t*)(bResult + 24))) < oWork.iTarget)
 					executor::inst()->push_event(ex_event(job_result(oWork.sJobID, results[i], bResult, iThreadNo, miner_algo), oWork.iPoolId));
 				else
--- a/xmrstak/backend/amd/minethd.hpp
+++ b/xmrstak/backend/amd/minethd.hpp
@@ -24,7 +24,7 @@ public:
 	static bool init_gpus();
 
 private:
-	typedef void (*cn_hash_fun)(const void*, size_t, void*, cryptonight_ctx*);
+	typedef void (*cn_hash_fun)(const void*, size_t, void*, cryptonight_ctx**);
 
 	minethd(miner_work& pWork, size_t iNo, GpuContext* ctx, const jconf::thd_cfg cfg);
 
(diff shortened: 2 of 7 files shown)


psychocrypt:
# Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz /4MiB cache) two hash one thread
- cryptonight 
  - dev: 117 H/S
  - PR: 119 H/S
- cryptonight_v7
  - dev: 117 H/S
  - PR: 117 H/S

I removed the digits after the comma because at least +-1 hash is the measurement error

psychocrypt:
I tested the PR on Windows, with NVIDIA and OpenCL miner. All works fine. The hash rate on different systems is better or equal to the old implementation.

psychocrypt:
I tested the macro's with clang, icc, msvc and gcc with different versions.

Spudz76:
I am excited about this but haven't had a chance to read through the new code, does it also allow for any number of low-power-mode threads now or is it still just 0-5?

Some CPUs of mine gained a good percentage from running 10 or 20 (and prefetch on) with the mbasaman hacks to the previous macros.

I will test this on all my CPUs and compilers too.  So far fastest for me are clang3.8 or gcc6 in most cases.

psychocrypt:
@Spudz76  It is now much simpler and needs less code to generate 10 or 20 hash versions. As a side effect the possibility to make mistake during the copy past is very low.

plavirudar:
I've tested on my CPUs that run N>2 scratchpads, and the performance was pretty unusual, in that the hashrate did not change linearly. 

CNL N=2 I get 150 on old and 155 on new implementation. 
CNL N=4 I get 250 on old and 245 on new implementation. 
CNL N=8 I get 200 on old and 170 on new implementation.

Spudz76:
[This diff from my fork](https://github.com/Spudz76/xmr-stak/commit/027b3f4635546b86887159dca5bf7ca2f4cc0aa2) contains the N=10 which I have been running for months, provides +30H/s on several lower TDP and cache starved CPUs.

Spudz76:
Also fun, with this new refactor you can try N=128 and while the code chooses N=1 the stats get confused and will show KH/s from each CPU core (not real of course).

The `iMultiway` should just get set to `MAX_N` if it is larger, before any of the rest of the logic.  And also set to 1 if lower than 1, and then the default in the case would be rendered ineffective (default being the cause of this bug).

PR #1800, opened 2018-08-22
Add systemd service unit file

aledomu:
Credits go to @codyps on the AUR, with a few tweaks of my own.

PR #1831, opened 2018-09-15
Monero POW v2

psychocrypt:
Implementation of the upcoming new Monero POW cryptonight_v8.

- add AMD/CPU and CUDA implementation
- add cpu single hash assembler version
- AMD: add option `unroll`
- CPU: add option `asm` to select assembler version of cryptonight_v8

Note: The CUDA implementation is currently slow and can be updated before the next release.

**ATTENTION** old config files fill not be compatible with this PR

Please do not report any hash reports here. For hash comparison please report it to the corresponding issue #1832

- [x] update docs: `pools.tpl`, `README.md` and  `amd.txt` (strided_index:1 is forbidden)
- [x] set dev pool to `monero8`
~~- [] add currency `monero8` and remove `monero7`~~ will be go in as soon as POW is final defined from the Monero devs
- [x] test NVIDIA `bfactor`
- [x] test https://github.com/fireice-uk/xmr-stak/pull/1831#issuecomment-421902099
- [x] fix https://github.com/fireice-uk/xmr-stak/pull/1831#issuecomment-422542145


xmrstak/backend/cpu/crypto/cryptonight_aesni.h
```diff
@@ -489,16 +556,27 @@ inline void cryptonight_monero_tweak(uint64_t* mem_out, __m128i tmp)
 	uint64_t idx0; \
 	__m128i bx0; \
 	uint8_t* l0 = ctx[n]->long_state; \
+	/* BEGIN cryptonight_monero_v8 variables */ \
+	__m128i bx1; \
+	__m128i division_result_xmm; \
+	__m128i sqrt_result_xmm; \
```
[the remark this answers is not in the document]
psychocrypt: thx for the info: I also saw it but have not looked deeper into the change to `uint64_t`. I will test it soon.

SChernykh:
@psychocrypt I'm going to make asm optimized versions for double hash today (single hash is done already), you should use them to get the best performance on CPU. It should be easy to integrate since I use old xmr-stak-cpu for testing.

psychocrypt:
thx I will have a look this evening and add also the unrolling as option to
the amd backend.

Spudz76:
I'm always four minutes too soon haha

Spudz76:
Instantly gives me issues on clang++-3.8 (others still attempting build):
```
[  3%] Building ASM object CMakeFiles/xmr-stak-asm.dir/xmrstak/backend/cpu/crypto/asm/cryptonigh_v8_main_loop.S.o
/usr/src/xmr-stak/xmrstak/backend/cpu/crypto/asm/cryptonigh_v8_main_loop_ivybridge.inc:160:2: error: invalid operand for instruction
 mov r13, -4389456576512
 ^
/usr/src/xmr-stak/xmrstak/backend/cpu/crypto/asm/cryptonigh_v8_main_loop_ivybridge.inc:167:2: error: invalid operand for instruction
 mov r13, 4389456576511
 ^
/usr/src/xmr-stak/xmrstak/backend/cpu/crypto/asm/cryptonigh_v8_main_loop_ryzen.inc:160:2: error: invalid operand for instruction
 mov rdx, 4389456576511
 ^
/usr/src/xmr-stak/xmrstak/backend/cpu/crypto/asm/cryptonigh_v8_main_loop_ryzen.inc:167:2: error: invalid operand for instruction
 mov rdx, -4389456576512
 ^
```

xmrstak/backend/cpu/crypto/cryptonight_aesni.h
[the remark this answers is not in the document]
psychocrypt: @SChernykh I tested it on an ryzen 1600X and the performance will go down a little bit if I switch to `uint64_t`, therefore this change is not included

Spudz76:
It will not build on this version of CMake (which is the debian stretch backports version `3.9.5-1~bpo9+1`)
crash caught with ALL compilers on this CMake version:
```
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/gcc-6
-- Configuring done
CMake Error at CMakeLists.txt:496 (add_library):
  Cannot find source file:

    xmrstak/backend/cpu/crypto/asm/cryptonigh_v8_main_loop.S

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
```

Using cmake-3.12.1 (debian sid package version `3.12.1-1`) on the one that works, but that is not what most users will be using, and is difficult to obtain on jessie/trusty or stretch/xenial and so on.  So I suggest fixing the extension (or extension search list) rather than bumping the supported CMake version at the top.

Spudz76:
This fixes the clang problem, by using 'gcc' regardless of anything else before enabling the ASM plugin.

I've also cleaned it up so the lib only gets linked when it is applicable, and added some x86_64 guards (in case ARM/neon ever gets added, etc)

[cmake-asm-improvement.patch.txt](https://github.com/fireice-uk/xmr-stak/files/2387086/cmake-asm-improvement.patch.txt)

Spudz76:
I fixed the ICC with gcc-7 backer problem by checking for an immutable macro that will appear when ICC is involved (even if `-no-icc` is passed to stealth itself, which would unset simpler macros such as `_ICC`)
```
-#if defined _MSC_VER || (__GNUC__ >= 7)
+#if defined _MSC_VER || ((__GNUC__ >= 7) && (!defined __INTEL_COMPILER_BUILD_DATE))
        _addcarry_u64(_subborrow_u64(0, x2, n0, (unsigned long long int*)&x2), r, 0, (unsigned long long int*)&r);
```

Thus, raw gcc-7 itself will still catch the intrinsics version, but gcc-7 as a backer for another toolchain will use the generic.

There appear to be lots of problems with [these particular intrinsics](https://stackoverflow.com/questions/33690791/producing-good-add-with-carry-code-from-clang?noredirect=1&lq=1) especially in clang, but it is strange that gcc-7 can consume them but not when it is being called via ICC.  Mayy be worth making a real assembly inline for each compiler to ensure it is not converted into garbage code smashing the stack and registers (which seems to be what every version of clang does with them).

psychocrypt:
@Spudz76 I fixed many things in my last two commits:

https://github.com/fireice-uk/xmr-stak/pull/1831/commits/a348d995cb3fa0fee8921ee7df38509123a6804b
- fix assembler code to pass the clang compiler
- CMake: set asm file language
- fix icc with gcc-7 compile issue with `_addcarry_u64`

https://github.com/fireice-uk/xmr-stak/pull/1831/commits/cf458d695e19c63728cb79dd0fc18f031f9528da
- Remove the asm option `auto` by `off`

psychocrypt:
The fix for the clang asm compile https://github.com/fireice-uk/xmr-stak/pull/1831/commits/a348d995cb3fa0fee8921ee7df38509123a6804b#diff-2fe31be299ace6e788b71e821d6b8877R160 for ryzen is currently not tested. I will do this this evening.

SChernykh:
> The fix for the clang asm compile a348d99#diff-2fe31be299ace6e788b71e821d6b8877R160 for ryzen is currently not tested.

- This will probably compile to the exact same machine code
- It's in sqrt_fixup branch which is executed only on 1 of ~470,000 iterations on average, so it shouldn't change performance at all.

xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
```diff
@@ -617,37 +664,96 @@ __kernel void JOIN(cn1,ALGO) (__global uint4 *Scratchpad, __global ulong *states
...
+			const uint d = (((uint *)c)[0] + (sqrt_result << 1)) | 0x80000001UL;
+ 			// Quotient may be as large as (2^64 - 1)/(2^31 + 1) = 8589934588 = 2^33 - 4
+			// We drop the highest bit to fit both quotient and remainder in 32 bits
+			division_result = fast_div_v2(RCP, c[1], d);
+ 			// Use division_result as an input for the square root to prevent parallel implementation in hardware
+			sqrt_result = fast_sqrt_v2(c[0] + as_ulong(division_result));
+#endif
+// cryptonight_monero_v8
+#if(ALGO==11)
+			{
```
SChernykh: This #if can be merged with the previous one.

xmrstak/backend/amd/amd_gpu/opencl/fast_int_math_v2.cl
```diff
@@ -0,0 +1,136 @@
...
+	const bool undershoot = (tmp >= b);
+
+	return (uint2)(
+		as_uint2(q).s0 + (undershoot ? 1U : 0U) - (overshoot ? 1U : 0U),
+		as_uint2(tmp).s0 + (overshoot ? b : 0U) - (undershoot ? b : 0U)
+	);
+}
+
+inline void fast_div_full_q(const __local uint *RCP, ulong a, uint b, ulong *q, uint *r)
+{
```
SChernykh: fast_div_full_q was there for testing only, it's not used in cn1(). It can be removed.

xmrstak/backend/amd/autoAdjust.hpp
```diff
@@ -127,6 +127,24 @@ class autoAdjust
...
+			// set strided index to default
+			ctx.stridedIndex = 1;
+
+			// nvidia performance is very bad if the scratchpad is not contiguous
+			if(ctx.isNVIDIA)
+				ctx.stridedIndex = 0;
+
+			// use chunked (4x16byte) scratchpad for all backends. Default `mem_chunk` is `2`
+			if(useCryptonight_v8)
+				ctx.stridedIndex = 2;
```
SChernykh: Our tests have shown that worksize=16 or 32 is optimal for the new algorithm depending on GPU, default worksize=8 is always slower.

xmrstak/backend/amd/jconf.cpp
```diff
@@ -150,6 +151,13 @@ bool jconf::GetThreadConfig(size_t id, thd_cfg &cfg)
 		return false;
 	}
 
+	if(!unroll->IsUint64() || (int)unroll->GetInt64() >= 128 )
+	{
+		printer::inst()->print_msg(L0, "ERROR: unroll must be smaller than 128");
+		return false;
+	}
+	cfg.unroll = (int)unroll->GetInt64();
```
SChernykh: You need to also check that unroll is a power of 2. OpenCL code will fail to compile otherwise.

xmrstak/backend/nvidia/nvcc_code/cuda_core.cu
```diff
@@ -194,6 +194,31 @@ __forceinline__ __device__ uint32_t shuffle(volatile uint32_t* ptr,const uint32_
...
+	uint64_t x = (n0 >> 12) + (1023ULL << 52);
+	const double xx = sqrt( *reinterpret_cast<double*>(&x) );
+	uint64_t r = *reinterpret_cast<const uint64_t*>(&xx);
+
+	const uint64_t s = r >> 20;
+	r >>= 19;
+
+	uint64_t x2 = (s - (1022ULL << 32)) * (r - s - (1022ULL << 32) + 1);
+
+ 	if (x2 < n0) ++r;
```
SChernykh: Leaving only one fixup step (r+=1) and removing the other (r-=1) is tricky - it requires specific FPU rounding mode to be set before the main loop. Not sure it's relevant as it will be replaced with fast sqrt which uses both fixup steps anyway.

psychocrypt:
@SChernykh is you asm code compiling under windows with MSVC? I got an error on the line `movq	 r13, -4389456576512` like `A2070 invalid instruction operants`

SChernykh:
@psychocrypt movq is only used with XMM registers: https://c9x.me/x86/html/file_module_x86_id_201.html

Was it clang that didn't like `mov r13, -4389456576512` ? It's 100% clang bug. You can try to workaround this (4389456576512 = 1022 << 32):
```
mov r13d, -1022
shl r13, 32
```

SChernykh:
Or you can `#define SET_REG(x, value) mov x, value` for MSVC and `#define SET_REG(x, value) movq x, value`

psychocrypt:
> Or you can `#define SET_REG(x, value) mov x, value` for MSVC and `#define SET_REG(x, value) movq x, value` for clang.

thx for the fast replay. Short after I asked I found that I have changed the white spaces and this was why MSVC was unhappy. Never the less a macro for mov is needed.

psychocrypt:
@SChernykh somehow MSVC not like if I add a define or macro into the àsm`file. Do you know if there is a special syntax for e.g. `#define MOVE mov`

SChernykh:
I think it's better to rewrite the code there than having a headache with macros.
-4389456576512 = -1022 << 32
4389456576511 = ~(-4389456576512) - use `not` instruction

SChernykh:
@psychocrypt Can you try these changes to `.inc` files: https://github.com/SChernykh/xmr-stak-cpu/commit/647fd8066ed2d78ef64f271f3c9eda2426b8f397 ? Performance is the same and you don't need to have the same code duplicated for Windows and Linux.

psychocrypt:
> @psychocrypt Can you try these changes to `.inc` files: [SChernykh/xmr-stak-cpu@647fd80](https://github.com/SChernykh/xmr-stak-cpu/commit/647fd8066ed2d78ef64f271f3c9eda2426b8f397) ? Performance is the same and you don't need to have the same code duplicated for Windows and Linux.

I have not tried this `inc` files but I can see that parts I changed to support msv compiler are still in. I have now a version for win64 and linux/bsd

SChernykh:
No, I mean, can you try these specific changes:
https://github.com/SChernykh/xmr-stak-cpu/commit/647fd8066ed2d78ef64f271f3c9eda2426b8f397#diff-2aabff7fe0697e74a96c50ce99193618R160
https://github.com/SChernykh/xmr-stak-cpu/commit/647fd8066ed2d78ef64f271f3c9eda2426b8f397#diff-c99bfd2e00aa22922f74f6d28d2f49b1R160
It's better because it doesn't require copy-paste for Windows and Linux.

psychocrypt:
> No, I mean, can you try these specific changes:
> [SChernykh/xmr-stak-cpu@647fd80#diff-2aabff7fe0697e74a96c50ce99193618R160](https://github.com/SChernykh/xmr-stak-cpu/commit/647fd8066ed2d78ef64f271f3c9eda2426b8f397#diff-2aabff7fe0697e74a96c50ce99193618R160)
> [SChernykh/xmr-stak-cpu@647fd80#diff-c99bfd2e00aa22922f74f6d28d2f49b1R160](https://github.com/SChernykh/xmr-stak-cpu/commit/647fd8066ed2d78ef64f271f3c9eda2426b8f397#diff-c99bfd2e00aa22922f74f6d28d2f49b1R160)
> It's better because it doesn't require copy-paste for Windows and Linux.

I will take it on my list. For now I keep it as it is and will move it to a follow up PR. I have it now running and must add your review comments and must work that I can check the checkboxes in the initial description.

side note: I know why I only develop for linux, windows is the pain for a developer.

Spudz76:
I also separately used the `movq` patch to make it compile those 64bit immediates,

I disagree movq is only for XMM regs, it is for any 64bit register.  Aren't XMM regs 128-bit anyway?

SChernykh:
Bug report: it crashes if "asm" is missing for any of CPU threads in cpu.txt.

fireice-uk:
> I disagree movq is only for XMM regs, it is for any 64bit register. Aren't XMM regs 128-bit anyway?

> some other description says the various MOV flavors only change the size of the operands (Addressing Modes section), nothing about type of register, other than there is no 128bit MOV and using MOVQ with XMM regs clears the upper 64 bits. So if MOVQ is just for XMM regs then it is severely underpowered for its only intent, haha.

It is also slower than reading the said XMM register from L1 cache @psychocrypt however fine optimisation can be left for later.

Spudz76:
However, plus side is the old non-asm version nicely ignores the existence of `asm` if you revert the executable and leave the `cpu.txt`

Crashes no matter what I try with v8 on CUDA 9.1 both via normal CUDA mode and using openclvendor switch.  Still runs v7 fine either version and both APIs about the same speed (~570H/s per 1060)

kio3i0j9024vkoenio:
This is the latest results with the most recent code for both:

https://github.com/fireice-uk/xmr-stak/pull/1831

and

https://github.com/SChernykh/xmr-stak-cpu/issues/1


Tests were done on the same hardware with the same configuration settings


OS: Ubuntu 16.04

Backend: (CPU Only)

    CPU: 4x Xeon E7-8837
    cryptonight_v7: 1598 H/s
    cryptonight_v8: 1332 H/s

Miner config:

"cpu_threads_conf" :
[
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 0 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 1 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 2 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 3 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 4 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 5 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 6 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 7 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 8 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 9 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 10 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 11 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 12 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 13 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 14 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 15 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 16 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 17 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 18 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 19 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 20 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 21 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 22 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 23 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 24 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 25 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 26 },
{ "low_power_mode" : true, "no_prefetch" : false, "affine_to_cpu" : 27 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 28 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 29 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 30 },
{ "low_power_mode" : false, "no_prefetch" : true, "affine_to_cpu" : 31 },
],

This 1332 H/s result is a little better than before but is still 16.6% worse than the 1596 H/s I get getting for the same v8 test with 

SChernykh/xmr-stak-cpu

https://github.com/SChernykh/xmr-stak-cpu/issues/1#issuecomment-422470749

asm_version" : 2
double hash prefetch turned on

Latest results (new code):
Base: 1656.6 H/S - 100%
INT_MATH: 1576.3 - 95.1%
SHUFFLE: 1609.0 H/S - 97.1%
BOTH: 1595.9 H/S - 96.3%

kio3i0j9024vkoenio:
> 
> 
> @kio3i0j9024vkoenio you need to post it here: #1832
> And you also need to add "asm" parameter to each CPU thread. Try both "asm" : "intel" and "asm" : "ryzen" to see what gives you best performance.

I am using "asm" : "ryzen" in the cpu config file.In the post above I just cut-n-pasted from an older post and forgot to show the latest config. Post above has been fixed.

"asm" : "ryzen" gave me the best result but again it was only 1332 H/s

SChernykh:
@kio3i0j9024vkoenio Double hash asm optimized version is not added to xmr-stak yet. And current double hash implementation in xmr-stak doesn't even use C++ optimized code: https://github.com/SChernykh/xmr-stak-cpu/blob/master/crypto/cryptonight_aesni.h#L488

So you won't get high hashrate with double threads now. But you can compare single hash only mode.

psychocrypt:
@fireice-uk  I checked all boxes in my PR. I would say the PR is mergeable. If there are additional changes e.g. performance need we should do it in a follow up. To increase this PR will not help.

Spudz76:
Assembly works on clang 3.8, 3.9, 4, but fails on clang 5, 6, 7, 8 with our old friend:
```
/usr/src/xmr-stak/xmrstak/backend/cpu/crypto/asm/cryptonight_v8_main_loop_ivybridge_linux.inc:160:2: error: invalid operand for instruction
 movq r13, -4389456576512
 ^
/usr/src/xmr-stak/xmrstak/backend/cpu/crypto/asm/cryptonight_v8_main_loop_ivybridge_linux.inc:167:2: error: invalid operand for instruction
 movq r13, 4389456576511
 ^
/usr/src/xmr-stak/xmrstak/backend/cpu/crypto/asm/cryptonight_v8_main_loop_ryzen_linux.inc:160:2: error: invalid operand for instruction
 movq rdx, 4389456576511
 ^
/usr/src/xmr-stak/xmrstak/backend/cpu/crypto/asm/cryptonight_v8_main_loop_ryzen_linux.inc:167:2: error: invalid operand for instruction
 movq rdx, -4389456576512
 ^
```

As the OSX Xcode compiler is either clang-6 or clang-8 based this should be fixed before merge.

I am using the official [llvm toolchain repo](http://apt.llvm.org/) for all the weird various versions of clang

SChernykh:
This should fix it for good: https://github.com/SChernykh/xmr-stak-cpu/commit/647fd8066ed2d78ef64f271f3c9eda2426b8f397#diff-2aabff7fe0697e74a96c50ce99193618R160

xmrstak/backend/amd/config.tpl
```diff
@@ -9,17 +9,20 @@ R"===(
  *                 2 = chunked memory, chunk size is controlled by 'mem_chunk'
  *                     required: intensity must be a multiple of worksize
  *                 1 or true  = use 16byte contiguous memory per thread, the next memory block has offset of intensity blocks
+ *                             (not allowed for cryptonight_v8 ans monero8)
```
Spudz76: replace `ans` with `and`

xmrstak/backend/nvidia/nvcc_code/cuda_core.cu
```diff
@@ -259,7 +296,7 @@ __global__ void cryptonight_core_gpu_phase2( int threads, int bfactor, int parti
 		for ( int x = 0; x < 2; ++x )
 		{
 			j = ( ( idx0 & MASK ) >> 2 ) + sub;
-
+
```
Spudz76: whitespace ghost

psychocrypt:
it is now late, I will fix it tomorrow night.

bomb-on:
@SChernykh Your changes in `cnv2_main_loop_ivybridge.inc` and `cnv2_main_loop_ryzen.inc` helped me to successfully build on macOS (10.13.6 with Apple LLVM version 10.0.0 (clang-1000.11.45.2)).

Spudz76:
Yep those asm changes worked it now compiles on everything I have.

And wow latest Xcode is clang-10 which I don't even see on the daily clang builds for debian or ubuntu - thanks for the report of it working there.

Repeating that I never got it to work in v8 mode on nVidia 1060.  Auto values put threads at "4141" which seems excessive :)  normal v7 values and v8 crashes, and same through opencl mode.

xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
```diff
@@ -617,37 +664,96 @@ __kernel void JOIN(cn1,ALGO) (__global uint4 *Scratchpad, __global ulong *states
...
+			uint index = ((b_x[0].s2 >> 26) & 12) | ((b_x[0].s2 >> 23) & 2);
+#	endif
+			b_x[0].s2 ^= ((table >> index) & 0x30U) << 24;
+			SCRATCHPAD_CHUNK(0) = b_x[0];
+			idx0 = c[0] & MASK;
+// cryptonight_monero_v8
+#elif(ALGO==11)
+			SCRATCHPAD_CHUNK(0) = b_x[0] ^ ((uint4 *)c)[0];
+#	ifdef __NV_CL_C_VERSION
+			// flush shuffeled data
```
Spudz76: 'shuffeled' should be 'shuffled'

xmrstak/backend/amd/amd_gpu/opencl/cryptonight.cl
```diff
@@ -658,44 +764,55 @@ __kernel void JOIN(cn1,ALGO) (__global uint4 *Scratchpad, __global ulong *states
...
 #endif
 
 			((uint4 *)a)[0] ^= tmp;
-			idx0 = a[0];
 
-			b_x = ((uint4 *)c)[0];
+// cryptonight_monero_v8
+#if (ALGO == 11)
+#	if defined(__NV_CL_C_VERSION)
+			// flush shuffeled data
```
Spudz76: 'shuffeled' should be 'shuffled'

Spudz76:
@psychocrypt I have applied most of what needed catching up [against your fork](https://github.com/psychocrypt/xmr-stak/compare/topic-cn-v8_v2...Spudz76:topic-cn-v8_v2?w=1)

psychocrypt:
could you please create a pr against my pr. but please without bittube
changes. Than I can merge it directly in my pr.

Tony Butler <notifications@github.com> schrieb am Mi., 19. Sep. 2018, 10:40:

Spudz76:
done

SChernykh:
@psychocrypt I hope this will help with your CUDA implementation: https://github.com/SChernykh/fast_int_math_v2_cuda - div and sqrt implemented and tested.

psychocrypt:
@SChernykh thx for your cuda work, I will have a look to it later.

I have now rebased the branch against the dev and added the asm fixes.

SChernykh:
@psychocrypt This code is literally a drop-in replacement, it tooks me a few minutes to get it up and running. I'm testing GTX 660 Ti now.

psychocrypt:
@SChernykh I added again some optimizations. The hash rate for may GTX1080 is now much better (with default configuration).

psychocrypt:
I will fore rebase this branch in a few min because I removed by accident some hash algos^^

psychocrypt:
OK now after 5 force push of the last two commits I repaired my mistake. The PR can now tested

Spudz76:
[Odd side effect of the new code](https://github.com/fireice-uk/xmr-stak/pull/1800#pullrequestreview-157109303), it doesn't work with "memory write execute" blocked, but the old version was fine.

aledomu:
This is what that option is blocking: <https://www.freedesktop.org/software/systemd/man/systemd.exec.html#MemoryDenyWriteExecute=>

I don't know much about ASM but I guess it's a side effect of doing dark wizardry to squeeze out the most performance possible.

I've deleted the line if you decide to go ahead, but I recommend if possible to figure out a way to not weaken the sandbox. Compromised miners are no fun.

SChernykh:
@psychocrypt There is "_TEXT_CNV8_MAINLOOP SEGMENT PAGE READ EXECUTE" which is compatible with "deny write execute" mode on Windows, but I don't see anything similar in cryptonight_v8_main_loop.S. I guess you need to specify segment attributes there too.

Spudz76:
Thanks @aledomu I use your .service file on several systems, but it must be Stretch/Xenial level else like half of the file has to be commented out (Jessie/Trusty have systemd too old to know most of those new tricks).  I guess what I'm saying is even an insecure .service file is better than none. :)  Friends don't let friends use `pm2` haha

aledomu:
Indeed I had to comment a few options in my Debian Stretch box. Since the sources are small enough for oneself to compile in a reasonable time, I thought I'd leave the task to comment those options to the user, who would have to deal with the sources anyway.

Not only Debian users need to tinker with the unit file: CentOS still uses an old-as-hell systemd version.

My goal with this unit file was to have it upstream so anybody who wanted a unit file had at least most of the work done. I was targeting mostly DIY rolling distros and systemd portable services, but I personally think that `docker`/`podman` is the best solution here in a more "stable" setup.

If there is some kind of policy for supporting old Linux distros, I will comment the required lines.

Spudz76:
Still can't get CUDA of any sort to work, tried Windows and it throws at [cuda_extra line 421](https://github.com/psychocrypt/xmr-stak/blob/topic-cn-v8_v2/xmrstak/backend/nvidia/nvcc_code/cuda_extra.cu#L421)
CUDA 9.2 said it was not autoconfiguring due to "only" 3842MB free.
CUDA 9.1 made a config that seemed ok but line 421 crash
Compiled CUDA 8.0 but didn't try it yet.  Generally have not had good luck with 8.0 clients to 9.2 driver even on old well proven code, and don't feel like reverting driver at this moment just for it to do likely the same (421 explosion).

Windows GTX970 via OpenCL does work, comparable rates in v7, but v8 runs around 3 times slower that way.  Assuming due to lack of nVidia-OpenCL optimal int-sqrt kernel so it uses normal long-hand float sqrt which I think we already knew would be slow.

Did not retest Linux yet but expect same as before (no nVidia via either method)

kio3i0j9024vkoenio:
Where exactly is the  fireice-uk:dev code?

There used to be a link to a zip of the v8 code to test but it seems to be missing now.

I would like to test the latest v8 code on my HP DL580 G7 server but need access to the code to do so.

PR #1834, opened 2018-09-17
avoid OpenCL binary missmatch

psychocrypt:
Avoid that a OpenCL binary from the cache is used if the driver or xmr-stak version has changed.


Spudz76:
Perfection!  This bit a lot of users and this makes it work like they expect.

PR #1839, opened 2018-09-19
configEditor: add version tagging, line filtering per-platform

Spudz76:
* Applies my recent changes from `FAQ.md` to the similar docs inlined with the main `config.txt`
* Add global `// generated by` header to all config files, to help keep things straight
* New feature allows for filtering template string lines per-platform, by placing a simple tag-string at the EOL of optional lines.  Untagged lines show up for all platforms (same as always)
* Immediately uses endtagging/filter feature in a few templates where it makes sense: Windows executable will drop the long documentary `config.txt` comment section about Linux sysctls, and vice versa.  Also optionalizes the platform-specific tips in `cpu.txt` to help signal to noise ratio
* Changes default `use_slow_memory` when Linux to `no_mlck` (keeps `warn` for Windows)


Spudz76:
Feel free to apply the other big patch first, I will gladly refactor this on top of it.

xmrstak/pools.tpl
```diff
@@ -22,7 +23,7 @@ POOLCONF],
  *
  *    aeon7 (use this for Aeon's new PoW)
  *    bbscoin (automatic switch with block version 3 to cryptonight_v7)
- *    bittube (uses cyrptonight_bittube2 algorithm)
+ *    bittube (uses cryptonight_bittube2 algorithm)
```
psychocrypt: could you please provide this as separate PR. We try to avoid different topics in one PR. This is than much better for referencing in case of issues and also writing change logs.

Spudz76: I applied it in the other patch under the Spelling Police so it can just be dropped here, for now

Depends which order you will be applying - I kept the other patchset atomic for whitespace / spelling / actual code separately

xmrstak/misc/configEditor.hpp
```diff
@@ -42,6 +43,24 @@ struct configEditor
 
 	void write(const std::string filename)
 	{
+		// endmarks: for filtering full lines inside the template string
+    // Platform marks are done globally here
```
psychocrypt: indention error

PR #1841, opened 2018-09-19
Add Intel ICC 18 detection/support to CMakeLists.txt

Spudz76:
* Detect ICC-18, add their header and lib paths

* Select newest available and compatible GCC and pass to ICC to avoid built-in default of using `/usr/bin/gcc` which in some distro revisions is too old
* Use `-DICC_GCCVER=N` to force specific  `gcc-N`

* Link to Intel lib


Spudz76:
Fresh on the heels of ICC-19 release, here is ICC-18 support...
I have not yet tried ICC-19

CMakeLists.txt
```diff
@@ -28,6 +28,25 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.1)
         message(FATAL_ERROR "g++ version must be at least 5.1!")
     endif()
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "Intel")
+    include_directories(BEFORE "/opt/intel/include")
```
psychocrypt: Please do not hard code any path's. It could be that the compiler is installed somewhere else even if this path is the default path.

CMakeLists.txt
```diff
@@ -28,6 +28,25 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
...
+    link_directories(BEFORE "/opt/intel/lib")
+    if((CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 17) AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 19))
+        # locate best gcc version to pass to icc-18 (gcc-8 doesn't work, for example)
+        # Use -DICC_GCCVER=6 for example to force 'gcc-6' and 'g++-6'
+        # Otherwise, the newest available and compatible pair will be selected
+        find_program(ICC_GXX NAMES "g++-${ICC_GCCVER}" "g++-7" "g++-6" "g++")
+        get_filename_component(ICC_GXX ${ICC_GXX} NAME)
+        find_program(ICC_GCC NAMES "gcc-${ICC_GCCVER}" "gcc-7" "gcc-6" "gcc")
+        get_filename_component(ICC_GCC ${ICC_GCC} NAME)
+        set(ICC_COMPILERS "-gxx-name=${ICC_GXX} -gcc-name=${ICC_GCC}")
```
psychocrypt: Not sure if this is the correct way, because if I install icc I select already an gcc version. This looks like a workaround.

Spudz76: What, where?  I run the installer, remove everything but the C++, it never asks it only seeks whatever `/usr/bin/gcc` points to... system default

I also use this to generate test binaries made with every permutation of ICC+GCC regardless what was configured or what the system defaults are.  So yes, it is a workaround, for having to use the ICC environment at all (I don't use their shell scripts or whatever).  I also don't have ICC in the system path.

Spudz76:
Example of what happens otherwise:
```
[ 31%] Building CXX object CMakeFiles/xmr-stak-backend.dir/xmrstak/backend/backendConnector.cpp.o
In file included from /usr/src/xmr-stak/./xmrstak/backend/pool_data.hpp(3),
                 from /usr/src/xmr-stak/./xmrstak/backend/miner_work.hpp(3),
                 from /usr/src/xmr-stak/./xmrstak/backend/globalStates.hpp(3),
                 from /usr/src/xmr-stak/xmrstak/backend/iBackend.hpp(3),
                 from /usr/src/xmr-stak/xmrstak/backend/backendConnector.cpp(24):
/usr/include/c++/8/cstdint(38): catastrophic error: cannot open source file "bits/c++config.h"
  #include <bits/c++config.h>
                             ^

compilation aborted for /usr/src/xmr-stak/xmrstak/backend/backendConnector.cpp (code 4)
```

It has picked up gcc-8 and crashes out.  You probably have never had gcc-8 on the same system as icc-18?  This compiler test box has gcc 5,6,7,8 and clang 3.8,3.9,4.0,5.0,6.0,7,8 and icc-18.

icc by itself is too stupid to avoid gcc-8 even though it does not work.

Spudz76:
Confirmed, I set my system compiler back to gcc-7 and icc works without these mods.
However if `/usr/bin/gcc` is actually `gcc-8` then icc crashes.

PR #1842, opened 2018-09-19
Adjust MSVC detection/support in CMakeLists.txt

Spudz76:
* Add `vim` modeline to end of file to help ensure correct indents (no tabs, 4 spaces)
* Patch whitespace where indent did not match above rules
* When MSVC, protect `-std=c++11` from being added to `CUDA_NVCC_FLAGS` (avoid warnings during compile)
* Add `/LTCG` to global linker flags (avoid "restarting linker" warning during link of openssl/hwloc deps)
* Add also `/INCREMENTAL:NO /OPT:REF` to shared linker flags for smaller EXE (unused references removal)
* Add `/Ox /GL /EHsc` to compiler flags to generate output compatible with above new linker flags


Spudz76:
I have been using these patches on my own "daily driver" rigs for over 6 months, pretty sure they work nice.  Mostly makes a handful of the pointless build warnings go away, does not appear to cause any problems.

psychocrypt:
Could you please remove the white space changes in this PR. Please do never mix formatting of a full file with any improvements. 
If you add a separate PR where all white spaces changes are in I can check it without doing any test. Also tehre is no need to mentioned white space changes in the change log but changing linker flags are important.

PR #1847, opened 2018-09-23
webdesign: Add backend thread tags to Thread IDs; Add pretty coin nam…

### LINKED ISSUE #1223: Wrong name in http daemon XMR-Stak Monero Miner
when i open localhost:8080
I see "XMR-Stak Monero Miner" in first black row.
But i mining not monero!
It must be or name algorithm, or current mining coin, but not always "Monero"

Spudz76:
* Add backend thread tags to Thread IDs; Someone once suggested but now I can't find link; Good idea anyway

* Add pretty coin names for page header; Fixes #1223

* Add extended cointype w/algo to results page

### DIFF
--- a/xmrstak/backend/cryptonight.hpp
+++ b/xmrstak/backend/cryptonight.hpp
@@ -2,7 +2,9 @@
 #include <stddef.h>
 #include <inttypes.h>
 #include <type_traits>
+#include <string>
 
+// NOTE: when changing this enum, check the cn_algo_name() below for any mismatch
 enum xmrstak_algo
 {
 	invalid_algo = 0,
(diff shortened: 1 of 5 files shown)


Spudz76:
Reworked first version as I decided it was done badly.

This one is much, much improved.

xmrstak/misc/executor.cpp
```diff
@@ -1036,9 +1037,26 @@ void executor::http_hashrate_report(std::string& out)
...
 	double fTotal[3] = { 0.0, 0.0, 0.0};
+	auto bTypePrev = static_cast<xmrstak::iBackend::BackendType>(0);
+	std::string name;
+	size_t j = 0;
 	for(size_t i=0; i < nthd; i++)
 	{
 		double fHps[3];
+		char csThreadTag[25];
+		auto bType = static_cast<xmrstak::iBackend::BackendType>(pvThreads->at(i)->backendType);
+		if(bTypePrev == bType){
```
psychocrypt: `{` must be on  a new line. Independent of the project where you provide code always look how the code style in the file looks like and try to follow this style (even if there is no style guide).

Spudz76: repaired

psychocrypt:
I will check this PR soon to have this PR in one of the next releases

Spudz76:
I have been running it on all my normal miners.
Wondered if this additional info might be better in a mouseover to keep clutter down but still provide the full data.  Then it could show other info like device name, PCI-slot-id (once I finish making that work for both gpu backends), etc.  Also working on NVML for monitoring since people have been wanting that.

psychocrypt:
note: NVML is not supporting all nvidia gpus.

Spudz76:
note: NVML is basically the only way to get/map the PCI slot info

note: nVidia likes to make everything difficult

Spudz76:
Yes, PCI location info should be available from everything

some pre-fermi may have not had NVML however we don't care about that

Spudz76:
clocks/temp/watts definitely spotty and available only on newer stuff in general
but pci slot info first really
other stuff because if NVML is there might as well harness it

Spudz76:
also been reading up on the AMD equivalent to make sure the backend connector can speak the same to both backends (abstracted to internal API)
not just based on how NVML does things / glue layer

xmrstak/backend/cryptonight.hpp
```diff
@@ -19,6 +21,60 @@ enum xmrstak_algo
 	cryptonight_monero_v8 = 11
 };
 
+/** get simplified/pretty coin name from xmrstak_algo value
+ *
+ * @param algo_id Any value, preferably from enum xmrstak_algo above unless you enjoy (unknown)
+ */
+inline std::string cn_algo_name(xmrstak_algo algo)
```
psychocrypt: Why is this function needed? We have already the selected currency name in jconf.cpp. We should use the names from there because translating an `algo` into a name of a currency is not possible. e.g. cryptonight_v8 is used by many currencies. We should always add the name from `pools.txt` into the webdialog and other reports.

Spudz76: Currency name from jconf is not "Human Pretty" but ugly lowercase whatever and less specific.
This is for the main Web header `XMR-Stak **WHATEVER** Miner`
It makes zero sense to show an ugly/algoname there thus needed another lookup table for the friendly names.

Regurgitating what they put in the config file also does nothing useful.

Could just change the header to `XMR-Stak CryptoNight Miner` and be done forever.  Mainly it shouldn't say Monero (unless user is actually mining Monero) nor should it be ugly, and it must be accurate.

Spudz76:
Updated for new added also (CN-SuperFast)

Spudz76:
I've split this into more atomic PR's (#2165 #2166 #2167) in hopes the acceptable parts can be included while the questionable bits can be improved or cancelled.

PR #1857, opened 2018-10-01
add CUDA 10.0 support

psychocrypt:
- extent MSVC workaround for CUDA to 10.0
- add compute architecture 75 if CUDA 10.0 is found


Spudz76:
Also, since `find_package(CUDA)` is deprecated and will never be updated again, it is safe to just use 9.0 or newer in the MSVC fix logic...

PR #1858, opened 2018-10-01
Dev

lordgh2018:
Exempt from 2 percent donations.


psychocrypt:
we are not ready for an release. We will have a release in ~1week

PR #1862, opened 2018-10-04
add cpu family and model detection

psychocrypt:
Helper functions to select the asm version based on the number of used hashes per threads and the family name of the cpu.

- use the newer cpu type functions to fix the wrong AMD family detection in `autoAdjust.hpp`
- allow to set the asm version to `auto`

Co-authored-by: fireice-uk <fireice-uk@users.noreply.github.com>


xmrstak/backend/cpu/config.tpl
```diff
@@ -12,7 +12,8 @@ R"===(// generated by XMRSTAK_VERSION
  *                  things slower.
  *
  * asm            - Allow to switch to a assembler version of cryptonight_v8; allowed value [off, intel, ryzen]
- *                    - off: used the default implementation (no assembler version)
+ *                    - auto: xmr-stak will automaticely detect the asm type (default)
```
Spudz76: `automaticely` should be `automatically`

PR #1871, opened 2018-10-07
OpenCL: fix definition range for unroll

### LINKED ISSUE #1870: `unroll` help text shouldn't include 0 as a valid option
The help text generated in `amd.txt` with the current `dev` branch says that the valid range for `unroll` is `[0;128)`, but 0 isn't isn't actually allowed (1 already means "don't unroll").  Trying 0 causes a compilation failure with:

```
/tmp/OCL112351T3.cl:1692:18: error: invalid value '0'; must be positive
                #pragma unroll CN_UNROLL
                               ^
<built-in>:9:19: note: expanded from here
#define CN_UNROLL 0
                  ^
1 error generated.
```

Perhaps the text here should be changed to read `[1,128)` instead?

psychocrypt:
fix #1870

- remove zero from the valid definition range for the loop unroll option


Bathmat:
Perhaps change either the bracket or parenthesis to match the other in the note too?
[Line 18](https://github.com/fireice-uk/xmr-stak/blob/746037d8fb33608224d6c2f17cbda91e5a328d3c/xmrstak/backend/amd/config.tpl#L18)

`(1;128)` or `[1;128]` ??

psychocrypt:
No the current one is the correct mathematician notation and means that 1
is included and 128 is excluded from the definition range.
https://en.m.wikipedia.org/wiki/Interval_(mathematics)

PR #1872, opened 2018-10-07
CPU: fix logical error

psychocrypt:
Fix wrong warning about unknown ASM type


psychocrypt:
I fixed all indents of the if (was totally wrong)

Issue #1880, opened 2018-10-09
Can we add some code so we can close the miner more gracefully?


caowenyu75:
Every time I want to close the miner, there is a chance that the computer might crash. Is there a way that this can be handled?

TIA

Please provide as much as possible information to reproduce the issue.

# Basic information
  - Type of the CPU.
  - Type of the GPU (if you try to miner with the GPU).
rx 550
# Compile issues
  - Which OS do you use?
  windows 10.
  add **all** commands you used and the **full** compile output here
  ```
  ```
  run `cmake -LA .` in the build folder and add the output here
  ```

# Issue with the execution
  - Do you compiled the miner by our own?
  yes and no.
  run `./xmr-stak --version-long` and add the output here
  ```

# AMD OpenCl issue

  ```
  run `clinfo` and add the output here
  ```

# Stability issue
  - Is the CPU or GPU overclocked? 
  - Is the Main memory of the CPU or GPU undervolted?

Spudz76:
set environment variable `CUDA_DEVICE_ORDER` to `PCI_BUS_ID` and then CUDA does what you want (bonus, matches nvidia-smi ordering)

PR #1885, opened 2018-10-09
clean memoy on miner shutdown

psychocrypt:
- free host and device memory on shutdown
- catch exceptions and try to do a clean shutdown (free mem)
- add key `q` to close the miner
- handle signals e.g. CTRL+C

solve #1880


xmrstak/misc/executor.cpp
```diff
@@ -997,6 +1001,52 @@ void executor::print_report(ex_event_name ev)
 	printer::inst()->print_str(out.c_str());
 }
 
+void executor::shutdown()
+{
+	// if shutdown is called the first time
```
fireice-uk: You are calling this from different threads. bQuit needs to be atomic. pvThreads is not safe. Use signals instead and wait conditions instead of calls and busy waiting.

psychocrypt:
no conditional is not possible you can not assume that all threads are
alive if the miner crash because of an exeption.

fireice-uk:
https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for

PR #1940, opened 2018-10-15
reduce blocking during metric update

psychocrypt:
With #1845 a race condition during the telemetry update is solved.
The problem is that the used mutex is blocking all threads from updating the metrics during
the statistics are calculated.

- introduce a mutex per miner thread


Spudz76:
I should have noticed that and used an array in the first place, thanks for repairing.

PR #1948, opened 2018-10-16
Add warnings for invalid results from CUDA and OpenCL

Dead2:
OpenCL and CUDA is generating hidden invalids, add warnings to make them visible. (Especially noticeable with newer AMD drivers where most results are invalids)


fireice-uk:
I think this one should be printed from the executor thread. Can you move this warning to `log_result_error` ?

fireice-uk:
They actually go onto report screen, but I agree that real-time notification will be useful too. I will close this one and move the print myself as we are trying to get the release out today.

Dead2:
Yep, the important thing is getting the warning into the next release :)

PR #1994, opened 2018-10-22
(doc/compile) Add better CUDA SDK vs Driver information

Spudz76:
Place detailed CUDA information in compile docs, to avoid repeating it 9729364 times manually in Issues


psychocrypt:
Could you olease link the source where it is announced that cuda 10 is
removing compute architecture 3.X?
I checked the release notes and the programming guide short after the
release and have not seen that any support is removed. Also compiling for
3.x works.

Spudz76:
No idea, guess it was a rumor.  Will remove that note

Spudz76:
All fixed up, should match reality.

PR #2003, opened 2018-10-24
NVIDIA: fix wrong number of threads

psychocrypt:
In the cuda backend for monero we start always twice as much threads as needed.
Those threads are than removed after the AES matrix is copied to the shared memory.
Never the less it is the result of an copy past bug.

- start correct number of threads for `monero`

### DIFF
--- a/xmrstak/backend/nvidia/nvcc_code/cuda_core.cu
+++ b/xmrstak/backend/nvidia/nvcc_code/cuda_core.cu
@@ -734,7 +734,7 @@ void cryptonight_core_gpu_hash(nvid_ctx* ctx, uint32_t nonce)
 {
 	dim3 grid( ctx->device_blocks );
 	dim3 block( ctx->device_threads );
-	dim3 block2( ctx->device_threads << 2 );
+	dim3 block2( ctx->device_threads << 1 );
 	dim3 block4( ctx->device_threads << 2 );
 	dim3 block8( ctx->device_threads << 3 );
 


Spudz76:
@xmrig brings Fermi to 75% perf of CN1 ("v7") instead of 50%

sergneo:
No, the hashrate fell, while    "threads" : 8, "blocks" : 128, for GTX 980  at 2.5.1 513 H / s 2.5.2 500H / s Monero v8

psychocrypt:
This is not possible if your setup has not changed. Before 2x the needed
threads are spawned. Now this is corrected.

sergneo:
Such poor results on 2.5.2 because I use compiled CUDA 8.0, before on CUDA 9.0 the results were quite bad, the hashrate was 1.5-2 times lower. It looks like this bug has been around for a long time, which you just now fixed in 2.5.2, now tested with CUDA 9.0 hashrate increased to 516 H/s

Spudz76:
CUDA 8.0 is non optimal for Maxwell, only really still supported for Fermi.
9.x is proper, 10.x also works very well.

CUDA 8.0 never worked very well via backward compatibility after CUDA 9.1 was in the drivers.  So compiled for CUDA 8.0 works best on last available CUDA 8.0-containing driver so ~386.28

Compile for whatever the driver contains always works best (check driver release note under supported technologies, it shows CUDA version bundled)  Backward compatibility is a crutch at best.

sergneo:
416.34 	CUDA - 10.0 
411.70 	CUDA - 10.0
411.63 	Added support for CUDA 10.0
399.24 	CUDA - 9.2
399.07 	CUDA - 9.2 
398.82	CUDA - 9.2
398.36	CUDA - 9.2
398.11 	CUDA - 9.2 
397.93	Added support for CUDA 9.2
397.64 	CUDA - 9.1
391.35	CUDA - 9.1 
391.24	CUDA - 9.1
391.01	CUDA - 9.1
390.77	CUDA - 9.1
390.65	CUDA - 9.1
388.71	Added support for CUDA 9.1
388.59	CUDA - 9.0
388.43	CUDA - 9.0
388.31 	CUDA - 9.0
388.13	CUDA - 9.0

Spudz76:
Also pending accept, new docs section [here](https://github.com/Spudz76/xmr-stak/blob/dev-cudaDocs/doc/compile.md#cuda-runtime-versus-cuda-sdk)

sergneo:
Tell me what version of CUDA optimal for the GTX 980 and GTX760 ?

sergneo:
xmr stack does not correctly detect the CUDA version of the Nvidia driver. With 388.13 miner writes at startup (9.1/9.0), 388.71 (9.2/9.0). A should have been (9.0/9.0) and (9.1/9.0) respectively.

psychocrypt:
are you sure. the miner is printing te value provided by the driver. If
cuda has no bug I think its correct

sergneo <notifications@github.com> schrieb am So., 28. Okt. 2018, 18:40:

Spudz76:
The release notes PDFs for 388.13/388.71 do claim such, however it must be a documentation lag as what the code says if from the driver itself so... 388.13 has 9.1 and 388.71 has 9.2 despite the docs.  There was the one other crossover noted in the new `compile.md` but I did not notice these two as I did not install every driver (table even says, based on PDFs, and even then for Windows since there are no standard release notes for the Linux ones)

The release notes PDF for 388.71 does say 9.1 however it must be a documentation lag as what the code says if from the driver itself so... 388.13 has 9.1 despite the docs.

sergneo:
Indeed, for example, driver 388.19 is installed with CUDA Toolkit 9.1. PDF is misleading.

PR #2001, opened 2018-10-23
Update for upcoming Graft CNv2 fork

jagerman:
Graft is forking to CNv2 (or "v8") in [just over a week](https://www.graft.network/2018/10/23/graft-anti-asic-major-network-update-1-5-on-october-31-2018/) with major version 11.  This PR updates the `graft` coin setting to recognize it.

PR #2011, opened 2018-10-25
bug fix release 2.5.2

psychocrypt:
# Changelog

- NVIDIA: fix wrong number of threads #2003
- Update for upcoming Graft CNv2 fork #2001


Spudz76:
Nice, went from 33.6 to 48 on old NVS 5200M (arch 20 / Fermi / laptop) - v7 rate was 64.

So, 50% fermi hit reduced to 75% pretty good for old hardware.

PR #2056, opened 2018-11-10
Patches to enable compilation under Windows with MSYS2 or Cygwin64

Spudz76:
Most features unsupported, use these cmake options: `-DCUDA_ENABLE=OFF -DOpenCL_ENABLE=OFF -DOpenSSL_ENABLE=OFF -DMICROHTTPD_ENABLE=OFF -DHWLOC_ENABLE=OFF`

However it does work, CPU only, no TLS... will work to improve further and get features to work properly.

Seems to magically exit when ASM mode is used, set `cpu.txt` asm:off

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,5 @@
-project(xmr-stak)
-
 cmake_minimum_required(VERSION 3.4.0)
+project(xmr-stak)
 
 # enforce C++11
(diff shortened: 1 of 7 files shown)


fireice-uk:
Your build fails CI - this needs to be fixed.

Spudz76:
Wow, this is old, I forgot about it.
Will update and repair CI

PR #2107, opened 2018-12-02
fix clamp implementation

psychocrypt:
Due to a wrong implementation clamp was not working.


xmrstak/backend/amd/config.tpl
```diff
@@ -43,6 +43,16 @@ GPUCONFIG
 ],
 
 /*
+<<<<<<< Updated upstream
+=======
```
fireice-uk: git fail =)

psychocrypt:
udated

PR #2110, opened 2018-12-02
OpenCL: auto config two threads per GPU

psychocrypt:
The auto config generates for AMD devices now by default two threads per GPU.

- remove the savety 128MiB memory now only from the max available GPU memory not from the avaialble memory for one alloc call
- extend the memory documentation in amd.txt


fireice-uk:
> extend the memory documentation in amd.txt

There seems to be no change like that.

PR #2115, opened 2018-12-02
fix default interleave value

psychocrypt:
The default value for interleave was wrongly set to 50.

Remove the value and take the default from the default constructor instead of side channeling it from the json parser.

### DIFF
--- a/xmrstak/backend/amd/jconf.cpp
+++ b/xmrstak/backend/amd/jconf.cpp
@@ -122,21 +122,22 @@ bool jconf::GetThreadConfig(size_t id, thd_cfg &cfg)
 		return false;
 
 	// interleave is optional
-	if(interleave == nullptr)
-		cfg.interleave = 50;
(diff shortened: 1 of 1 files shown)


xmrstak/backend/amd/jconf.cpp
```diff
@@ -122,21 +122,22 @@ bool jconf::GetThreadConfig(size_t id, thd_cfg &cfg)
...
-	}
-	else
-	{
-		cfg.interleave = (int)interleave->GetInt64();
+		if(!interleave->IsUint64())
+		{
+			printer::inst()->print_msg(L0, "ERROR: interleave must be a number");
+			return false;
+		}
+		else if((int)interleave->GetInt64() < 0 || (int)interleave->GetInt64() > 100)
```
fireice-uk: It you want to have a plain int here you can simply just use `GetInt()`

PR #2165, opened 2019-01-09
(web) Add backend thread tags to Thread IDs on Hashrate Report

Spudz76:
Split from #1847


psychocrypt:
sry for the late response. Somehow I missed all the PRs. I will check it as soon as possible.

PR #2177, opened 2019-01-19
Add Support For Pool Supplied Block Versions

brandonlehmann:
This PR adds support for pool supplied Block versions in the miner job that is provided by the pool.

This alleviates some of the issues experienced with some projects that operate with a parent/root block and the majorVersion of that block is not updated during a fork for various reasons.

See https://github.com/turtlecoin/node-turtle-pool/commit/5671ef9a9c71fda7243815ef1be84087d4da6443 for an example of how the information can be supplied by a pool.

### DIFF
--- a/xmrstak/backend/miner_work.hpp
+++ b/xmrstak/backend/miner_work.hpp
@@ -21,12 +21,14 @@ namespace xmrstak
 		bool        bNiceHash;
 		bool        bStall;
 		size_t      iPoolId;
+		uint32_t    bMajorVersion;
+		uint32_t    bMinorVersion;
 
-		miner_work() : iWorkSize(0), bNiceHash(false), bStall(true), iPoolId(invalid_pool_id) { }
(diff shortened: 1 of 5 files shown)


brandonlehmann:
How is passing the version numbers ambiguous? 

Passing the algo from the pool in the job is far more likely to cause issues as each miner package is free to label an algo as they see fit including small variations, numbering them, or doing whatever they feel like. 

The only way I see sending the algo in a tag making sense is if someone operates a CN algo registry of sorts that all the miner packages look to for parameters. 

This, as implemented, is designed to provide an optional hint of what the miner should be using. It still relies on the miner software understanding what the block version means. It’s not definitive but it’s better than referring to the parent/root block only and/or relying on people to update their configs at just the right time. 

In any event, we’re using it downstream with success. If you don’t want to include it, feel free to close the pull.

fireice-uk:
@brandonlehmann please see jpsock.cpp for the registry of values - we have it since 2.0

brandonlehmann:
@fireice-uk I see your registry in passing the aglo and as stated before, if you don't feel this is worthy of inclusion, you're free to close the PR.

PR #2192, opened 2019-02-01
add ryo algorithm fork

psychocrypt:
switch with block 6 to cryptonigh_gpu
(https://github.com/ryo-currency/ryo-currency/blob/7bc5dc40864e518753807d04c17d617f728da156/src/cryptonote_config.h#L185)


LeadManPL:
> switch with block 6

I assume this is some typo? Current block is some around 226162 , right?

psychocrypt:
6 is the block version. Ryo is running currently version 5 and will fork to
version 6. The miner itself is stupid and knows nothing about the
blockchain.

LeadManPL:
ok, thanks for the explanation. Now how about 14th date, I didn't see any option in 2.8.0 miner configs to switch to new blockchain version. Will it switch automatically?

psychocrypt:
yes if you select ryo as currency the miner will switch if the pool provide
a work with the new block version

PR #2197, opened 2019-02-02
OpenCL: fix Blake hashing

psychocrypt:
Windows driver creates wrong code if unroll is used.


lkelemen:
"Windows driver creates wrong code if unroll is used."
Can I suppose that this means that there is no this kind of bug in the AMD driver of Mac-s?

psychocrypt:
I do not know. I have not get reports that mac was not working.

lkelemen:
I tried it on Mac and 2.8.0 still makes about 15% bad hashes. Can it be that the Mac and the PC gets the same hash 10-15% of the time and as the PC is faster the Mac will make bad hashes?

psychocrypt:
I do not understand you. The issue has nothing to do with the performance.

lkelemen:
OK

PR #2206, opened 2019-02-04
fix missing algorithm name

psychocrypt:
During the implementation of `cryptonight_gpu` we missed to add the name
of the algorithm to jpsock. 
This fix solves a incompatibility with xmrig-proxy.

- add function to translate an algorithm into a name.

reported by @xmrig


fireice-uk:
@xmrig Thanks - we both can barely see after January.

PR #2209, opened 2019-02-05
refactor POW definition

psychocrypt:
A POW is now defined by a function `f` and three degrees of freedom `f(iteration, scratchpad, mask)`.
`f` is the base algorithm like `cryptonight, cryptonight_gpu`

This change make it very easy to integrate the new trend of variate the
number of iterations or the scratchpad size without modifying the full
code.

- [x] tests required before merging


psychocrypt:
offline discussed with @fireice-uk : I will remove the long definition for the user and will use this mechanism only intern.

PR #2216, opened 2019-02-08
OpenCL: fix groestl again

psychocrypt:
With #2212 I fixed groestl for the windows driver 19.2.1 and VEGA gpus.
This fix break the rocm support.
Therefore I will switch to `unroll 3` to avoid hopfully both bugs.
@xmrig mentioned in a private chat the `unroll 3` is also solving the
windows issue.


psychocrypt:
replaced by #2219

PR #2226, opened 2019-02-10
update algorithm documentation

psychocrypt:
- add more information if a algorithm derives from an other
- update `pools.tpl`

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,11 +64,13 @@ If your prefered coin is not listed, you can choose one of the following algorit
     - cryptonight_lite_v7_xor (algorithm used by ipbc)
 - 2MiB scratchpad memory
     - cryptonight
+    - cryptonight_gpu (for Ryo's 14th of Feb fork)
     - cryptonight_masari (used in 2018)
     - cryptonight_v7
     - cryptonight_v7_stellite
     - cryptonight_v8
     - cryptonight_v8_half (used by masari and stellite)
+    - cryptonight_v8_zelerius
 - 4MiB scratchpad memory
     - cryptonight_haven
     - cryptonight_heavy
--- a/xmrstak/pools.tpl
+++ b/xmrstak/pools.tpl
@@ -45,9 +45,12 @@ POOLCONF],
  *    cryptonight_lite_v7_xor (algorithm used by ipbc)
  *    # 2MiB scratchpad memory
  *    cryptonight
+ *    cryptonight_gpu (for Ryo's 14th of Feb fork)
  *    cryptonight_superfast
  *    cryptonight_v7
  *    cryptonight_v8
+ *    cryptonight_v8_half (used by masari and stellite)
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -57,18 +57,20 @@ Monero. See [here](https://github.com/fireice-uk/cryptonote-speedup-demo/) for d
 
 If your prefered coin is not listed, you can choose one of the following algorithms:
 - 256Kib scratchpad memory
-    - cryptonight_turtle
+    - cryptonight_turtle (equal to cryptonight_v8 but `0x1FFF0` as mask, 1/8 of iterations and scratchpad size)
```
fireice-uk: This is way too much information for the main readme - do we really need this?

README.md
```diff
@@ -57,18 +57,20 @@ Monero. See [here](https://github.com/fireice-uk/cryptonote-speedup-demo/) for d
...
 - 1MiB scratchpad memory
     - cryptonight_lite
     - cryptonight_lite_v7
     - cryptonight_lite_v7_xor (algorithm used by ipbc)
 - 2MiB scratchpad memory
     - cryptonight
+    - cryptonight_gpu (for Ryo 02/2019 fork)
```
fireice-uk: Let's make it `14th of Feb fork` - remember they write the dates opposite way on the other side of the pond.

README.md
```diff
@@ -57,18 +57,20 @@ Monero. See [here](https://github.com/fireice-uk/cryptonote-speedup-demo/) for d
...
     - cryptonight_masari (used in 2018)
     - cryptonight_v7
     - cryptonight_v7_stellite
     - cryptonight_v8
-    - cryptonight_v8_half (used by masari and stellite)
+    - cryptonight_v8_half (used by masari and stellite; equal to cryptonight_v8 but with half iterations)
+    - cryptonight_v8_zelerius (equal to cryptonight_v8 but with 393216 iteration)
```
fireice-uk: Same here. Algo is not a coin. If they can't be bothered to come up with original algo let's just call it `cryptonight_v8_half`. I don't think annotations are needed - just internal mappings will be enough

README.md
Spudz76: I prefer `YYYY-MM(-DD)` because everywhere across every pond does it wrong.
Significance ordering is always the right order, then it sorts properly as text also, and easy to read and comprehend/compare quickly without lexical errors.

README.md
psychocrypt: `cryptonight_v8_half` is an other algorithm with the half iteration count of v8. This algo uses an different iteration count.

README.md
fireice-uk: Exactly the reason why I prefer letters -> no matter the order "14th of Feb" "Feb 14th", the meaning is the same to everyone.

fireice-uk:
Documentation only - so skipping CI

PR #2240, opened 2019-02-14
[WIP] Increase max job input length (needed by Masari)

### LINKED ISSUE #2233: cl_invalid_kernel_args clenqueuendrangekernel
I get loads on these after the program starts running ...
Error CL_INVALID_KERNEL_ARGS when calling clEnqueueNDRangeKernel for kernel 0
I'm using crypto drivers from AMD + driver patcher for modded bios
Everything with the program itself is default ...

# Basic information
  - Type of the CPU. --noCPU
  - Type of the GPU (if you try to miner with the GPU). 4x rx480
  - I'm mining masari with the latest miner

# Compile issues
  - Which OS do you use?
 windows 10

# Issue with the execution
 Just downloaded the ZIP 

# AMD OpenCl issue
  - I think that the issue is here somewhere

# Stability issue
  - GPU sits on 2%

Do i have to install something else on my machine because it's a clean install of win10 ?
I can run the xmrig-amd-2.10.0 with no errors at all ... but wanted something newer ...

jagerman:
The current 84 maximum job length for jobs on AMD isn't sufficient for Masari, which now needs 108.  This increases it and also adds an error message to catch this if it happens again.

Fixes #2233

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/gpu.cpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.cpp
@@ -335,7 +335,7 @@ size_t InitOpenCLGpu(cl_context opencl_ctx, GpuContext* ctx, const char* source_
 		return ERR_OCL_API;
 	}
 
-	ctx->InputBuffer = clCreateBuffer(opencl_ctx, CL_MEM_READ_ONLY, 88, NULL, &ret);
+	ctx->InputBuffer = clCreateBuffer(opencl_ctx, CL_MEM_READ_ONLY, amd_max_input_len + 4, NULL, &ret);
 	if(ret != CL_SUCCESS)
 	{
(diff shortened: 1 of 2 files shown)


jagerman:
Apparently this isn't enough; it avoids the error, but is giving `AMD Invalid Result GPU ID 0`.

psychocrypt:
I saw it also that masari uses a not syd conform size. what is the reason
for that. It is not enough to increase the number on the host side since
all arrays in on device must also be extendet.

jagerman:
> I saw it also that masari uses a not syd conform size.  what is the reason for that. It is not enough to increase the number on the host side since all arrays in on device must also be extendet.

Yeah, I was far too quick on the gun submitting this PR; I'm going to close it.

There is some discussion of the change here: https://github.com/xmrig/xmrig/issues/913

PR #2268, opened 2019-03-03
cryptonight_r

psychocrypt:
This PR introduce cryptonight_r. The CPU version is currently not using ASM code and is therefore not showing the full performance.


ykutlu:
Hi,
please add UPX (uplexa)
Thanks...

PR #2273, opened 2019-03-04
OpenCl: fix multi gpu support

psychocrypt:
Since #2268 the multi gpu support is broken. It is not possible to have more than one opencl context in a process.
Use one OpenCl context for the process instead per GPU.


Ethorsen:
Fix does seem to work.  Can now start 2.9 properly

fireice-uk:
@Ethorsen Thanks for testing!

bdunford:
I also had the same issue and this change addressed it.   Thanks for all the great work.

PR #2274, opened 2019-03-04
CN_GPU: Fix old compilers and old kernels (Clang 3.5 tested, without CONFIG_HUGETLBFS)

Spudz76:
This makes CN_GPU compile on Clang 3.5, and then furthermore allow allocations without HUGETLB (with warning) since it is faster than always-slow mode even though not ideal.


psychocrypt:
thx for your patch. I will have a deeper loom to in in two days since I am
currently on a trip

PR #2277, opened 2019-03-06
Add pause, resume and quit support

cppdev-123:
This is a simple pull to provide pause, resume and quit for cpu threads up till now

I should have added the features to gpu threads as well but the laptob has gone for about two weeks and I'm using a mobile phone with github web editor but it's really a headache and very difficult to edit on github with a mobile phone

also I fear the original repo is updated before I pull so that I will have to reedit again


cppdev-123:
I apologise I forgot to add size_t num member

cppdev-123:
Why arebthe commits refused ?

fireice-uk:
Hi! The red X is CI build status. This means the code fails to run on some platform. Click the X find out more (keep in mind that if you push some code on top while the build runs it will fail because it will be cancelled)

fireice-uk:
You will also need to squash and clean up this PR

cppdev-123:
Do you mean fail to run or fail to build ?
I know there is some mistakes in code that prevent build because I have no computer at the current time but I'm still working

Also it's better to notify me with the mistakes in the code because sometimes I don't figure out what's wrong with it (clicking on X mark doesn't do anything)

fireice-uk:
It fails to build, clicking on the red x will give you the CI menu -> here is the list for your most recent build https://ci.appveyor.com/project/fireice-uk/xmr-stak/builds/22879803

Issue #2263, opened 2019-02-27
Please add new custom algorithm CryptoNight Heavy X 


AnimusXCASH:
Greetings,
I am from X-Cash team. I would like to kindly ask you if you can add our new custom algorithm to your mining software.
Proof Of Work algorithm was changed from Cryptonight V8 to modified Cryptonight HeavyX, to try to prevent ASICs and or FPGAs from mining on the network. This means as well that we are now nicehash resistant.

Daemon: https://github.com/X-CASH-official/X-CASH/releases/tag/1.5.0

You can reach us through the Discord community if any further information required:
https://discord.gg/sr5G3ch

Thank you for your time.
brg

Lovro
Marketing and Community Manager
X-Cash Team

Spudz76:
Would suggest not using keyword "Heavy" unless also 4MB scratchpad size (that is what most would consider what "Heavy" means with CN algos) it would only lead to more confusion.

2MB scratchpads are not Lite (1MB) nor Heavy (4MB)

If you double the work maybe try "Hard2X" or something

AnimusXCASH:
thank you for suggestion

psychocrypt:
What are the changes of x-cash pow against v8 or native cryptonight.
Please provide a pool address and a wallet address which I can use for testing.

psychocrypt:
added in #2316

PR #2316, opened 2019-03-10
add cryptonight_v8_double

psychocrypt:
close: https://github.com/fireice-uk/xmr-stak/issues/2263

Add new algorithm alias for cryptonight_v8_double. The difference to v8
is that the number of iterations is increased.

Add also the currency X-CASH which use cryptonight_double.


AnimusXCASH:
Hi,
mining pool
http://minexcash.com/#getting_started

The difference from our algo to regular CN V8 is we have double the iterations than CN V8. the scratchpad is still 2MB

Where do you want that I send you wallet?

psychocrypt:
Thx I already found the online wallet generator. I needed only a address
for the short tests against a pool. Since the pow only changed the number
of iterations it was easy to integrate it.

AnimusXCASH:
> Thx I already found the online wallet generator. I needed only a address
> for the short tests against a pool. Since the pow only changed the number
> of iterations it was easy to integrate it.

great, could you give as an EST. time when it will be integrated? 

thanks

psychocrypt:
my personal schedule is to release in the next 24h.

PR #2332, opened 2019-03-12
add cli option `amdCacheDir`

psychocrypt:
add option to control where the amd cache data should be stored

e.g.
`./xmr-stak -C supportxmr.txt --amdCacheDir $HOME/xxx`

### DIFF
--- a/xmrstak/backend/amd/amd_gpu/gpu.cpp
+++ b/xmrstak/backend/amd/amd_gpu/gpu.cpp
@@ -48,23 +48,10 @@
 
 #ifdef _WIN32
 #include <windows.h>
-#include <Shlobj.h>
 
 static inline void create_directory(std::string dirname)
 {
     _mkdir(dirname.data());
-}
-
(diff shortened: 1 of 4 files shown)


Kroch:
Hi!
sorry for offtop but what version of AMD driver can you recomend for Windows\Linux platforms for best performance?
Thanks!

Spudz76:
Depends much upon which type of GPU
Linux ROCm latest for anything newer than Polaris or so
Linux 17.xx for old GCN3 Islands cards or worse/older
Linux 15.xx for ancient evergreen series or early GCN
and similarly
Windows Adrenaline 18.xx for Vega
Windows Crimson 17.xx for R9/RX
Windows Omega 15.xx for evergreen or early GCN

but that's still just a general guide, and in the end you just have to test them on your setup
results are not identical for everyone (different mobo, different cpu, different ram, different... else - too many variables to say which is best-for-you)

PR #2341, opened 2019-03-14
Add new Algorithm for cryptonight_armor

2acoin:
Add new algorithm for cryptonight_armor
 * Add new currency for 2ACoin

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ XMR-Stak is a universal Stratum pool miner. This miner supports CPUs, AMD and NV
 
 Besides [Monero](https://getmonero.org), following coins can be mined using this miner:
 
+- [2ACoin](https://www.2acoin.org)
 - [Aeon](http://www.aeon.cash)
 - [BBSCoin](https://www.bbscoin.xyz)
 - [BitTube](https://coin.bit.tube/)
@@ -59,6 +60,7 @@ Monero. See [here](https://github.com/fireice-uk/cryptonote-speedup-demo/) for d
 
 If your prefered coin is not listed, you can choose one of the following algorithms:
 - 256Kib scratchpad memory
+    - cryptonight_armor
     - cryptonight_turtle
 - 1MiB scratchpad memory
     - cryptonight_lite
--- a/xmrstak/backend/cryptonight.hpp
+++ b/xmrstak/backend/cryptonight.hpp
@@ -28,6 +28,7 @@ enum xmrstak_algo_id
 	cryptonight_r = 16,
 	cryptonight_v8_reversewaltz = 17, //equal to cryptonight_monero_v8 but with 3/4 iterations and reversed shuffle operation
 
+	cryptonight_armor = start_derived_algo_id,
 	cryptonight_turtle = start_derived_algo_id,
 	cryptonight_v8_half = (start_derived_algo_id + 1),
 	cryptonight_v8_zelerius = (start_derived_algo_id + 2),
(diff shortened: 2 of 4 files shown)


fireice-uk:
Sorry but I can't see any difference between your algo and cn_turtle

psychocrypt:
Please use `cryptonight_turtle` to mine your coin. The coin should have at least 8MH/s global hashrate before we will add it.

psychocrypt:
cn_turtle is 4 times faster than e.g. Monero. The network hash of the
suggeated coin is currently 200kh/s this coresponse to 50kh/s monero
hashrate. Since it is possible still possible to mine even if the coin is
not supported nativly there is no need to support it.

timetoseek:
> Please use `cryptonight_turtle` to mine your coin. The coin should have at least 8MH/s global hashrate before we will add it.

Thank you, I was curious at which point you might support some of the new ones...  so a newish coin like GRIN (at only ~2MH/s), is nowhere on the horizon :frowning:

Spudz76:
Newish coins should send their programmers over here, probably well before their dumb coin launches, and add their dumb coin to the code themselves (via PR) to support their coin, if they wanted it supported at launch because nobody else is going to do it for them until the coin is popular.

This is how Open Source works?  It is not there to do all your work for you upon request.  Also stealing code technically, by not submitting the changes upstream.

Most new coins fork a miner, terribly hack the code like they don't know what they are doing, and then launch off of that and never properly clean up their hacks and submit an acceptable PR.  It's not really that hard to match a projects code style rules, and "do it right", before submitting a patch.  We also won't clean up terrible code, it can be lightly messy, but not completely nasty laughable hacks.  Most forked new-coin hacks don't actually add a new algo (properly) but destroy an existing one (thus the code can't be upstreamed) and probably other problems I can't recall right now.

When a coin is some silly minor change like number of rounds or whatever (CN-half, CN-double) then those are easier to add.  When they have some new tweak other than that, it is much more difficult.

Anyway that's why new coins get no love from here (and probably all other mainstream miner apps).  Otherwise we'd all be adding abandonware-coins constantly while most of them die off.  Whoever starts a coin should be the only one to lose a bunch of time supporting *their* coin.

timetoseek:
> Newish coins should send their programmers over here, probably well before their dumb coin launches, and add their dumb coin to the code themselves (via PR) to support their coin, if they wanted it supported at launch because nobody else is going to do it for them until the coin is popular.
> 
> This is how Open Source works? It is not there to do all your work for you upon request. Also stealing code technically, by not submitting the changes upstream.
> 
> Most new coins fork a miner, terribly hack the code like they don't know what they are doing, and then launch off of that and never properly clean up their hacks and submit an acceptable PR. It's not really that hard to match a projects code style rules, and "do it right", before submitting a patch. We also won't clean up terrible code, it can be lightly messy, but not completely nasty laughable hacks. Most forked new-coin hacks don't actually add a new algo (properly) but destroy an existing one (thus the code can't be upstreamed) and probably other problems I can't recall right now.
> 
> When a coin is some silly minor change like number of rounds or whatever (CN-half, CN-double) then those are easier to add. When they have some new tweak other than that, it is much more difficult.
> 
> Anyway that's why new coins get no love from here (and probably all other mainstream miner apps). Otherwise we'd all be adding abandonware-coins constantly while most of them die off. Whoever starts a coin should be the only one to lose a bunch of time supporting _their_ coin.

I agree :100:%, and thank you for such a succinct overview (and XMR-Stak is awesome!!) :clap:

PR #2362, opened 2019-03-20
[WIP] Optimized soft aes for windows and some other features

cppdev-123:
This should make the miner get the hashrate it had in before CryptonightR fork
but it currently works only on windows and crashes on linux and maybe osx
any help will be welcome


psychocrypt:
Thank you for you work. 
Please cleanup your pull request and change only files which are necessary for this feature. 
Somehow you touched 6800 lines.

cppdev-123:
the only files changed for soft aes are the asm files in cnR dir and cryptonight_aesni.h and CryptonightR_gen.cpp

some new asm files for msvc and gcc soft aes for cnR and wow were added so the there is a lot of lines edited

psychocrypt:
please have a look to the diff. I see many more changes. Also a lot of
whitespace changes.
Please really change only lines those are required to be changes.
Please also squash the commits together.

Spudz76:
moved to #2372
I did the above for you...
the other hacks look maybe cool but need similar treatment (and separation from this issue)

PR #2366, opened 2019-03-22
AMD: workaround driver memory leaks

psychocrypt:
Instead of creating a new kernel for each block we pre-build 25 kernel within a chunk.
This reduces the calls to `clCreateProgramWithSource` which creates the leaks and will crash the miner after a few hours.

Workaround was first introduced by @SChernykh in xmrig-amd

- add helper: `search_program()`
- split OpenCL runtime code into definition and body for `cryptonight_r`


psychocrypt:
yes it should take 25 times linger before it crashes. I have heard that the
latest driver works well

PR #2385, opened 2019-03-29
fix simple start

psychocrypt:
Simple starts requires always user interactions if config.txt and
pools.txt is not available.
This PR asked the user only on the first advanced option if he/she
preferse the simple start or an advanced mode.
If all parameter are already defined via command line options than the
question will never be shown.


vanities:
Thank you!

q629988171:
good job

PR #2386, opened 2019-03-29
avoid hash rate drop during dev pool mining

psychocrypt:
Since the dev pool is sometimes using another POW than the user pool the
hash date can strongly increase or decrease.
In a case of a hash rate drop a monitoring software could trigger a
wrong miner restart.

- use the hash rate calculated during the user pool mining to
interpulate the number of hashes during the dev pool mining so that the
statistics stay constant.


psychocrypt:
do not open PRs against the master^^

PR #2412, opened 2019-04-25
Fix CN_MASK

catalanscoin:

### DIFF
--- a/xmrstak/backend/cryptonight.hpp
+++ b/xmrstak/backend/cryptonight.hpp
@@ -183,7 +183,7 @@ struct xmrstak_algo
 // default cryptonight
 constexpr size_t CN_MEMORY = 2 * 1024 * 1024;
 constexpr uint32_t CN_ITER = 0x80000;
-constexpr uint32_t CN_MASK = ((CN_MEMORY - 1) / 16) * 16;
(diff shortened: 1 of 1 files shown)


psychocrypt:
why do you think that the current implementation is wrong.

fireice-uk:
@psychocrypt Both ways are exactly the same, read it as:

1. Subtract 1 to get all 1's
2. Shift right by 4 bits
3. Shift left by 4 bits

Whereas the new version is 

1. Shift right by 4 bits
2. Subtract 1 to get all 1's
3. Shift left by 4 bits

I prefer the first version since it is more organised logically.

psychocrypt:
@fireice-uk  You are right. thanks , that is what I missed^^

psychocrypt:
@catalanscoin Thanks for the PR but we will stay with the current implementation.

catalanscoin:
It is not the same.

"CN_MASK = ((CN_MEMORY - 1) / 16) * 16;" === "CN_MASK = CN_MEMORY - 1;"

The correct operation is this one: constexpr uint32_t CN_MASK = (CN_MEMORY / 16 - 1) * 16;

I think this const is not used anywhere, but I would fix It.

PR #2414, opened 2019-04-27
Fix CN_MASK

catalanscoin:
"CN_MASK = ((CN_MEMORY - 1) / 16) * 16;" == "CN_MASK = CN_MEMORY - 1;"
Current implementation only does -1. 16/16 is 1 so code is doing nothing.

The correct operation is this one: constexpr uint32_t CN_MASK = (CN_MEMORY / 16 - 1) * 16;

I think this const is not used anywhere, but I would fix It.


fireice-uk:
Computer science is not maths :) If you want to know what / 16 * 16  it is zeroing the lower 4 bits of the mask.

fireice-uk:
The const is actually the most critical part of the code. I will close it as a duplicate of @2412

PR #2415, opened 2019-04-27
--log CLI parameter

o2genum:
The `--log` CLI parameter lets the user start logging right away, instead of editing the config and starting the miner again.


psychocrypt:
thx for the PR

PR #2416, opened 2019-04-28
Don't fail with "use_slow_memory" == "warn" and --noUAC on Windows 7, 8

o2genum:
These conditions combined make the miner terminate:

- `"use_slow_memory"` == `"warn"`
- `--noUAC`
- Windows 7 or 8

[The code](https://github.com/fireice-uk/xmr-stak/blob/a5b8fb741b17307871681ef0eab6fd4cf224b6df/xmrstak/cli/cli-miner.cpp#L769) where the termination happens. This breaks the expected behavior of the `warn` option on Win 7 and 8,


fireice-uk:
Approved and thanks for debugging this one!

PR #2422, opened 2019-05-01
Multiple pools via CLI

o2genum:
Supply multiple pools via CLI options by using pool-related options multiple times.

Better CLI options functionality makes xmr-stak easier to integrate, because command line parameters are generally easier to manage programmatically.

### DIFF
--- a/xmrstak/misc/executor.cpp
+++ b/xmrstak/misc/executor.cpp
@@ -524,14 +524,21 @@ void executor::ex_main()
 	telem = new xmrstak::telemetry(pvThreads->size());
 
 	set_timestamp();
-	size_t pc = jconf::inst()->GetPoolCount();
 	bool dev_tls = true;
-	bool already_have_cli_pool = false;
-	size_t i = 0;
-	for(; i < pc; i++)
-	{
+	
+	size_t pc = jconf::inst()->GetPoolCount();
+	std::vector<jconf::pool_cfg> pool_cfgs;
+	for (size_t i = 0; i < pc; i++) {
 		jconf::pool_cfg cfg;
 		jconf::inst()->GetPoolConfig(i, cfg);
+		pool_cfgs.push_back(cfg);
+	}
+
+	// Add all pools from the config file.
+	// If a pool is also specified in CLI params, CLI values override config values.
+	for(size_t i = 0; i < pc; i++)
+	{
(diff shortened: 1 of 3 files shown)


xmrstak/misc/executor.cpp
```diff
@@ -524,32 +531,38 @@ void executor::ex_main()
 		if(!cfg.tls)
 			dev_tls = false;
 
-		if(!xmrstak::params::inst().poolURL.empty() && xmrstak::params::inst().poolURL == cfg.sPoolAddr)
+		auto& pool_from_params = find_if(xmrstak::params::inst().pools.begin(), xmrstak::params::inst().pools.end(), [&cfg](const xmrstak::pool &pool_from_params) { return pool_from_params.poolURL == cfg.sPoolAddr; });
```
psychocrypt: This line produces an error in linux: 

```
xmrstak/misc/executor.cpp:534:35: error: invalid initialization of non-const reference of type ‘__gnu_cxx::__normal_iterator<xmrstak::pool*, std::vector<xmrstak::pool> >&’ from an rvalue of type ‘__gnu_cxx::__normal_iterator<xmrstak::pool*, std::vector<xmrstak::pool> >’
   auto& pool_from_params = find_if(xmrstak::params::inst().pools.begin(), xmrstak::params::inst().pools.end(), [&cfg](const xmrstak::pool &pool_from_params) { return pool_from_params.poolURL == cfg.sPoolAddr; });
                            ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

psychocrypt:
I will move this PR release after the next, since I need first check the error.

xmrstak/misc/executor.cpp
mtl1979: When using "auto" as variable type, compiler doesn't automatically use "const" qualifier, so in this case it should be "const auto&" for the line to be valid. It is always best to explicitly add "const" qualifier if the following code is not supposed to modify the variable contents, this will allow compiler to optimize the code better.

PR #2424, opened 2019-05-01
Topic rename coin stellite

psychocrypt:

I have not changed the internal name `cryptonight_stellite` for the old POW of stellite.
If I need a procrastination task I will remove older cryptonight variants those are not longer used by the coins we support in a separate PR.


psychocrypt:
will reopen again because the CI has issues.

PR #2430, opened 2019-05-03
Fallback from hwloc autoconf to basic autoconf

o2genum:
If hwloc autoconf does not work, fall back to the basic autoconf.

This provides a better guess for some CPUs for which hwloc autoconf fails. This was marked as "TODO" in the code.


psychocrypt:
thx for this feature

PR #2431, opened 2019-05-03
--h-print-time CLI option

o2genum:
Useful for automated starts without having to edit the config and restart miner.


psychocrypt:
Thx I need to check if this works smoothly together with the simple configuration

PR #2432, opened 2019-05-03
Enabling/disabling GPUs from CLI

o2genum:
New CLI arguments for enabling/disabling specific GPUs:
```
--amdGpus GPUS             indices of AMD GPUs to use. Example: 0,2,3
--nvidiaGpus GPUS          indices of NVIDIAGPUs to use.
```

Disabled GPUs are commented out in nvidia.txt or amd.txt.

This is useful for automated starts on specific GPUs without config editing.


psychocrypt:
Thx I need to have a deeper look into it to avoid side effects

PR #2443, opened 2019-05-31
NVIDIA: optimize cryptonight_gpu

psychocrypt:
- optimize keccak
- avoid bank conflicts (shared memory) in phase3
- use `__byte_perm` for rot operation if maxwell or newer

increase cn-gpu speed by 10%


mhmdyj5949:
خیلی ممنون

PR #2494, opened 2019-07-31
AMD: optimize VEGA auto suggestion

psychocrypt:
- increase the work size to 16 for cryptonight_gpu for vega GPUs


golipormehdi:
س

PR #2495, opened 2019-07-31
NVIDIA: fix broken phase 3

psychocrypt:
- with #2468 I fully break the miner: producing invalid results and not scan all shares

The release 2.10.6 is fully broken for NVIDIA and should not be used.


AudirBiscacio:
Nao tem um link para que eu possa acessar diretamente??

psychocrypt:
in english please.

PR #2539, opened 2019-10-26
Add support for Talleo's mining algorithm.

mtl1979:
Talleo is rebranded Bittorium, which uses similar algorithm as TurtleCoin but without the "Lite" mask. No generic algorithm works with Talleo.

### DIFF
--- a/xmrstak/backend/cpu/minethd.cpp
+++ b/xmrstak/backend/cpu/minethd.cpp
@@ -465,6 +465,16 @@ bool minethd::self_test()
 			ctx[0]->hash_fn("This is a test This is a test This is a test", 44, out, ctx, algo);
 			bResult = bResult && memcmp(out, "\x30\x5f\x66\xfe\xbb\xf3\x60\x0e\xda\xbb\x60\xf7\xf1\xc9\xb9\x0a\x3a\xe8\x5a\x31\xd4\x76\xca\x38\x1d\x56\x18\xa6\xc6\x27\x60\xd7", 32) == 0;
 		}
+		else if(algo == POW(cryptonight_talleo))
+		{
+			func_selector(ctx, ::jconf::inst()->HaveHardwareAes(), false, algo);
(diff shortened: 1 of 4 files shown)


psychocrypt:
do you tested the changes with all three platforms cpu/amd/nvidia?

mtl1979:
I tested with cpu and nvidia... Don't have amd card.

psychocrypt:
can you please post a real pool and a corresponding wallet address

psychocrypt:
I need the pool also to check for the hashrate. we normally not accept
coins under 5mh/s (exceptions are sometimes possible)

mtl1979:
With 60 second target time, the hash rate won't climb very high... I made temporary build so our miners can use GPUs to mine, otherwise the network hash rate doesn't start climbing.

I used pool.talleo.org:3334 to verify my patch is correct.

My own wallet address is:

TA4yACzMYuFYq7V6xVAWYHeS39jQ8w4mKRowpY6NskGuS1rZpjcWuCpdeCypwUCJrK9mGqVW9o1pY2EG3HW7BZkR2YRcc4YNa

leisegang:
5mhs is now at the official pool. Can you please add talleo to xmr stak? 
https://leisegang.no/src/viphGPz7.png screenshot of hashrate

Pool port has changed to 3333
@psychocrypt

mtl1979:
On some AMD GPUs, the auto configuration seems to set the intensity to double of usable value. I'm not really experienced at fixing that part of the code...

psychocrypt:
the hashrate is very small let me discuss this with @fireice_uk. This will
take some time because we are busy with other development

mtl1979:
@psychocrypt We lost some miners because of a bug in mining pool software... We already fixed the bug on two pools and will fix it on the third too. It takes more time on the third pool because it is running nodejs-pool and not software designed specifically for Talleo.

PR #2551, opened 2019-11-16
Correct authors of RandomX mining code

SChernykh:
Since this branch offers only RandomX algorithm, I put correct author names for CPU/GPU mining code.


psychocrypt:
Thanks for the PR. I missed to change that part.

psychocrypt:
The ci not passed but the reason is yhat we need at least cuda 9. I will fix the CI. This PR contains only documentarion changes so ci can be ignored ;-)

Issue #2582, opened 2019-11-30
MEMORY ALLOC FAILED: VirtualAlloc failed


entak:
Since switch to randomx, I see the MEMORY ALLOC FAILED: VirtualAlloc failed for each working cpu thread. HW is the same (was working fine - without the error - before the switch), cpu config auto-generated.  I googled a bit and done following:
-run miner as admin
-checked that locking pages in Local Security Policy is allowed for the account
-restarted the system, run the mining as a very first thing
-increased Virtual RAM Size (I did not find any tips how much to increase it, so tried to set it to 40GB virtual RAM on machine with 8GB physical RAM, 5 mining threads)

None of the above helped. Any tips? Or logs where can I find closer info about the issue? Thx in advance.

breadsax:
I have the same error, but only on machine with AMD CPU. Intel CPU does not have this error.

entak:
Few more observations:
-I do have it on Intel CPU as well - i5 8500, AMD is Ryzen 3900X
-lowering number of mining threads has no effect - tried with single thread
-running miner under Win10

pershoot:
I am getting this (MEMORY ALLOC FAILED: VirtualAlloc failed) on GPU (RX560) on an AMD CPU host (FX-8350).

psychocrypt:
how do you enabled large page support. Try to run `sudo sysctl -w vm.nr_hugepages=1280` maybe your number of large pages is to low!

entak:
I am using Win OS, so through the Local Security Policy. I didn't find any option to set large pages count there ...

diacomi:
Same problem here, but I noticed that occurs only if I use --noTest parameter.

psychocrypt:
can you please post the output with and without --noTest

diacomi:
Sure...

**WITH --noTest**

xmr-stak-rx.exe --noTest
-------------------------------------------------------------------
xmr-stak-rx 1.0.1-rx 7a9361b22

Brought to you by fireice_uk and psychocrypt under GPLv3.
Based on CPU mining code by tevador and SChernykh.
-------------------------------------------------------------------
You can use following keys to display reports:
'h' - hashrate
'r' - results
'c' - connection
-------------------------------------------------------------------
   #####   ______               ____
 ##     ## | ___ \             /  _ \
#    _    #| |_/ /_   _   ___  | / \/ _   _  _ _  _ _  ___  _ __    ___  _   _
[...]

PR #2624, opened 2019-12-09
[RX] fix windows huge page support

### LINKED ISSUE #2598: Windows MEMORY ALLOC FAILED: VirtualAlloc failed if running with noTest
I built windows binary with affinity fix from https://github.com/fireice-uk/xmr-stak/pull/2590
But if I run miner with --noTest I got many of MEMORY ALLOC FAILED: VirtualAlloc failed

It affects hashrate, 2500 with error and 2900 without error

Rig: 2xE5-2660v2, 2x4GB RAM, Windows 10 1803 build 17134.1069

psychocrypt:
fix #2598 and #2582

If the command line option `--noTest` was used the huge pages for
windows was not activated.

### DIFF
--- a/xmrstak/backend/cpu/minethd.cpp
+++ b/xmrstak/backend/cpu/minethd.cpp
@@ -260,6 +260,12 @@ bool minethd::self_test()
 	if(res == 0 && fatal)
 		return false;
 
+	if(!params::inst().selfTest)
+	{
+		printer::inst()->print_msg(L0, "skip self test: disabled by the command line option '--noTest')");
+		return true;
+	}
+
 	cryptonight_ctx* ctx[MAX_N] = {0};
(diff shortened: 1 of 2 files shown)


fireice-uk (approved):

dchmelik:
On this topic, seems windows huge page support isn't well-documented nor reported by the miner whether it's using it or not (unlike others like cpuminer.)  It's good GNU/Linux xmr-stak-rx is more-documented, but I have to use it on windows for a while as am also using teamredminer (GPU) which temporarily had problems in GNU/Linux.  I'd like the configuration to be documented in config.txt if necessary and xmr-stak-rx to say whether it's using huge pages or not, and what types, or have a visible menu option to display that.

PR #2673, opened 2020-05-06
[RX] release 1.0.5-rx

fireice-uk:
### Changelog
- Disable GPU support for RandomX by default
- Add keva, safex support
- Bring xmr-stak-rx up to date with newest xmrig
#2671


lyrenhexfan1:
Idiot copy paster kid

Issue #2685, opened 2020-08-16
error: invalid conversion from ‘int (xmr-stak-1.0.5-rx)


Ultracaine:
`...` are the placeholder for your answers. Please answer each question!

**Describe the bug**

**Which operating system do you use? **

```
System:    Host: Ryzen Kernel: 5.7.14-1-MANJARO x86_64 bits: 64 compiler: gcc v: 10.1.0 Desktop: KDE Plasma 5.19.4 
           Distro: Manjaro Linux 
Machine:   Type: Desktop Mobo: ASUSTeK model: PRIME X370-PRO v: Rev X.0x serial: <superuser/root required> 
           UEFI: American Megatrends v: 5603 date: 07/28/2020 
Battery:   ID-1: hidpp_battery_0 charge: N/A condition: N/A model: Logitech Gaming Mouse G900 status: Charging 
CPU:       Topology: 8-Core model: AMD Ryzen 7 2700X bits: 64 type: MT MCP arch: Zen+ rev: 2 L2 cache: 4096 KiB 
           flags: avx avx2 lm nx pae sse sse2 sse3 sse4_1 sse4_2 sse4a ssse3 svm bogomips: 118223 
           Speed: 2805 MHz min/max: 2200/3700 MHz Core speeds (MHz): 1: 2924 2: 1713 3: 1984 4: 1862 5: 1903 6: 2180 7: 2196 
           8: 2196 9: 1792 10: 1716 11: 1884 12: 2254 13: 2195 14: 2198 15: 2537 16: 1991 
Graphics:  Device-1: Advanced Micro Devices [AMD/ATI] Vega 10 XL/XT [Radeon RX Vega 56/64] vendor: ASUSTeK driver: amdgpu 
           v: kernel bus ID: 0c:00.0 
           Display: x11 server: X.Org 1.20.8 driver: amdgpu FAILED: ati unloaded: modesetting s-res: 3440x1440 
           OpenGL: renderer: Radeon RX Vega (VEGA10 DRM 3.37.0 5.7.14-1-MANJARO LLVM 10.0.1) v: 4.6 Mesa 20.1.5 
           direct render: Yes 
Audio:     Device-1: Advanced Micro Devices [AMD/ATI] Vega 10 HDMI Audio [Radeon Vega 56/64] driver: snd_hda_intel v: kernel 
           bus ID: 0c:00.1 
[...]

PR #2688, opened 2020-09-04
Fix: "invalid conversion from ‘int" when compiling version 1.0.5 (microhttpd 0.9.71)

DGdev91:
Had issue compiling on ArchLinux, i currently have microhttpd 0.9.71.
Seems like i'm not the only one, there's an open issue complaining for the same problem #2685

Using the MHD_Result enum instead of "int" solved the problem (still had to make a static cast for MHD_digest_auth_check, which returns an int)


DGdev91:
The pach was intended for xmr-stak-rx, i made a pull request for the right branch. https://github.com/fireice-uk/xmr-stak/pull/2689

But actually the code looks the same, probably is useful on the main branch too

PR #2689, opened 2020-09-04
Fix: "invalid conversion from ‘int" when compiling version 1.0.5 (mic…

DGdev91:
Had issue compiling on ArchLinux, i currently have microhttpd 0.9.71.
Seems like i'm not the only one, there's an open issue complaining for the same problem #2685

Using the MHD_Result enum instead of "int" solved the problem (still had to make a static cast for MHD_digest_auth_check, which returns an int)

### DIFF
--- a/xmrstak/http/httpd.cpp
+++ b/xmrstak/http/httpd.cpp
@@ -46,7 +46,7 @@ httpd::httpd()
 {
 }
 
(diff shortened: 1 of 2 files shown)


superzerg:
thanks for the fix, I had the same problem on Debian 10 and it worked like a charm !

